PR #18, opened 2019-02-10
Replace bower with webpack

nick-nikoliuk:


webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+      }, {
+        test: /\.(gif|png|jpe?g|svg)$/i,
+        use: [{
+          loader: 'file-loader'
+        }],
+      }
+    ]
+  },
+  mode: 'production',
+  amd: {jQuery: true},
```
DarthVanger: `jQuery` is not used in this app. Do we need `jQuery: true` in this case?

webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+  entry: './src/index.js',
+  output: {
+    filename: 'index.js',
+    publicPath: ASSET_PATH,
+    path: path.resolve(__dirname, 'dist')
+  },
+  module: {
+    rules: [
+      {
+        //   use: 'babel-loader',
```
DarthVanger: what is this comment for?

webpack.config.js
```diff
@@ -0,0 +1,71 @@
...
+        use: [
+          {
+            loader: "style-loader" // creates style nodes from JS strings
+          },
+          {
+            loader: "css-loader" // translates CSS into CommonJS
+          }
+        ]
+      }, {
+        test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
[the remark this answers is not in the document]
nick-nikoliuk: No we don't. Were added for future versions. Removed.

webpack.config.js
nick-nikoliuk: Accordingly documentation, better to use it flag in case any 3rd party library using jQuery or will use in feature.
https://github.com/amdjs/amdjs-api/wiki/jQuery-and-AMD

webpack.config.js
nick-nikoliuk: For future going to ES6, I saw it in issues.

PR #22, opened 2019-04-06
Refactor components using React

nick-nikoliuk:


src/ui/app-root.js
```diff
@@ -1,50 +1,45 @@
...
-      break;
-    case 'notes':
-      renderNotesPage();
-      break;
-    default:
-      throw new Error('unknown page: ', page);
+import {NotesPage} from 'ui/notes-page';
+
+export const PAGES_ENUM = {
+  EMPTY: Symbol('empty'),
```
DarthVanger: What is an empty page? Why is it needed?

src/ui/login/login.js
```diff
@@ -1,73 +1,110 @@
...
+    gapiAuthorize()
+      .then(() => {
+        console.debug('login.js: auth success! calling thoughtStorage.scanDrive()');
+        siteGlobalLoadingBar.hide(spinnerName);
+      })
+      .then(googleDriveApi.loadDriveApi)
+      .then(thoughtStorage.scanDrive)
+      .then(() => {
+        console.info('login: drive scanned, redirecting to the app main page');
+        this.props.authorized();
```
DarthVanger: I'd call it something like `onLoginSuccess` or `onAuthSuccess` or `handleAuthSuccess`.

`authorized` is quite ambiguous: one might think it's a boolean or a function which returns a boolean.

src/ui/login/login.js
```diff
@@ -1,73 +1,110 @@
...
 import googleApiLoader from 'api/google-api-loader';
 import './slide-1.png';
 import './slide-2.png';
 import './slide-3.png';
 import './slide-4.png';
 import './login.css';
 
-let element;
+export class LoginPage extends Component {
+  state = {googleApiInit: false};
```
DarthVanger: I'd call it `isGoogleApiInitialized`.

`googleApiInit` sounds more like a function which would initialize the google api.

src/thought/view-thoughts/thoughts-mind-map-view.js
```diff
@@ -1,281 +1,286 @@
...
+        id: targetThoughtId
+      }
+    });
+    brainVisNetwork.selectNote(targetThoughtId);
+  }
 
-  currentViewedThought = {
-    id: node.id,
-    name: node.label
+  onChangeClick = name => {
```
DarthVanger: I think this is `onNoteSelect` or `onNoteSwitch` or `handleNoteSwitch`.

`onChangeClick` sounds quite weird

src/thought/view-thoughts/thoughts-mind-map-view.js
```diff
@@ -1,281 +1,286 @@
...
-  currentViewedThoughtId = targetThoughtId;
-  thoughtStorage.logTree();
-  var targetThought = thoughtStorage.findThoughtById(targetThoughtId);
+    window.open(thoughtStorage.getLinkToThought(this.state.currentViewedThought));
+  };
 
-  if (!targetThought) {
-    throw new Error('changeThought(): couldn\'t find targetThought in thoughtStorage by id: ', targetThoughtId);
+  closeThoughtNameEditor() {
+    this.setState({editNote: null});
```
DarthVanger: I think this should be something like `isNoteBeingEdited: boolean` or `noteMode: view | edit` or `isInNoteEditMode: boolean`.

src/thought/view-thoughts/thoughts-mind-map-view.js
```diff
@@ -1,281 +1,286 @@
...
+  render() {
+    const {editNote} = this.state;
 
-  brainVisNetwork = new BrainVisNetwork({
-    containerDomElement: element
-  });
+    return (
+      <div ref={this.ref} id="thoughts-container">
+        {editNote && <ThoughtNameEditor
+          thought={thoughtStorage.findThoughtById(editNote.id)}
```
DarthVanger: editing is always for a selected note, so we could use `selectedNote.id` here.

Or if we want a separate variable for this, we could name it `noteToEdit` or `noteBeingEdited`.

src/thought/view-thoughts/thoughts-mind-map-view.js
```diff
@@ -1,281 +1,286 @@
...
+    if (!thought.isNote) {
+      console.info('It is not allowed to edit user-uploaded files');
+      return;
+    }
 
-  var node = brainVisNetwork.visNodes.get(currentViewedThoughtId_temp);
-  console.debug('node from visNodes: ', node);
+    this.setState({
+      currentViewedThought: thought,
+      editNote: {
```
DarthVanger: `editNote` is basically the same thing as the name of the method: `editThought` (after rename it will be `editNote()`).

So this should be `editedNote` or `noteToEdit` or `editorNote` or `noteForEditor`. And actually it's the same as selected note, so maybe makes sense to just use `selectedNote` here in order to simplify the naming problem.

src/ui/app-root.js
nick-nikoliuk: When the app starts, data for app is still not prepared. So we using empty page while app decides to change it to login or main screen.

src/ui/login/login.js
nick-nikoliuk: done

src/ui/login/login.js
nick-nikoliuk: done

src/ui/app-root.js
nick-nikoliuk: It will be removed after sagas implementation

src/thought/view-thoughts/thoughts-mind-map-view.js
nick-nikoliuk: вщту

src/thought/view-thoughts/thoughts-mind-map-view.js
nick-nikoliuk: changed to "selectedNote", bacause it's object, not a boolean

src/thought/view-thoughts/thoughts-mind-map-view.js
nick-nikoliuk: renamed to "selectedNote"

src/thought/view-thoughts/thoughts-mind-map-view.js
nick-nikoliuk: renamed to selectedNote

PR #25, opened 2019-05-11
Implement App composite component (reducer, container, component, sagas)

nick-nikoliuk:


src/components/App/AppContainer.js
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
+import {AppComponent} from 'components/App/AppComponent';
```
DarthVanger: is linter configured to enforce no space or a single space between the brackets for the `import { }` statement?

webpack.config.js
```diff
@@ -6,12 +6,13 @@ const CopyWebpackPlugin = require('copy-webpack-plugin');
 const ASSET_PATH = process.env.ASSET_PATH || '/';
 
 module.exports = {
-  entry: './src/index.js',
+  entry: ['@babel/polyfill', './src/index.js'] ,
```
DarthVanger: what does this do?

src/components/Spinner/SpinnerActions.js
```diff
@@ -0,0 +1,2 @@
+export const SHOW_SPINNER_ACTION = 'SHOW_SPINNER_ACTION';
```
DarthVanger: This action is used only in `SpinnerSagas.js`. The saga exports `showSpinner()`.

So basically this action is internal for the saga, nobody is going to ever reuse it, because only `showSpinner()` is for external use.

We're going to end up with `SpinnerActions.js` being used only in `SpinnerSagas.js`. And the file has only 2 lines and I guess we're not going to add more actions for the spinner ever?

On the other looks very well structured. Maybe the structure is more important than having no 2-line files :)

If we had one big file for all the actions it would be hard to refactor (split) it later.

show-spinner-action

src/sagas.js
```diff
@@ -0,0 +1,24 @@
...
+  applyMiddleware(sagaMiddleware)
+);
+
+sagaMiddleware.run(rootSaga);
+
+export const action = (type, data = null) => store.dispatch({type, data});
+
+export function* rootSaga() {
+  yield all([
+    signIn(),
```
DarthVanger: Not sure `sagas.js` is the right place for calling `signIn()`.

Except this line, all the surrounding code is a boilerplate config for the saga library itself.

But `signIn()` is more about starting the application. The first action that is dispatched.

Or is it obvious that `rootSaga()` in `sagas.js` is the place where app starts execution? Sorry didn't read the docs yet :)

src/components/App/AppSagas.js
```diff
@@ -0,0 +1,33 @@
...
+import thoughtStorage from 'storage/thought-storage';
+import googleDriveApi from 'api/google-drive-api';
+import googleApiLoader from 'api/google-api-loader';
+import {hideSpinner, showSpinner} from 'components/Spinner/SpinnerSagas';
+
+function setPageAction(data) {
+  return put({type: CHANGE_PAGE_ACTION, data});
+}
+
+export function* loadApp() {
```
DarthVanger: I would assume that `loadApp` is where the app starts loading, makes first requests etc.

But in `rootSaga()` we call `signIn()`, not `loadApp()`. So are we beginning the app with `signIn()` instead of beginning it with `loadApp()`? I'm confused.

src/sagas.js
```diff
@@ -0,0 +1,24 @@
...
+    watchSpinner()
```
DarthVanger: Will we need to initilize all watchers here, for each action like showing a spinner?

webpack.config.js
nick-nikoliuk: Because sagas using generators that still not implemented in all browsers. So we using polyfills for that.

src/components/Spinner/SpinnerActions.js
nick-nikoliuk: I agree with that. In this case this file a redundant and was created to avoid change structure in future. But you're right. For now it can be removed.

src/sagas.js
nick-nikoliuk: Yes, this is the place where we subscribe saga for action. Not for every action but for every saga file. I will change that naming.

src/components/App/AppSagas.js
nick-nikoliuk: That is naming issue. I will try to make it more clear.

src/sagas.js
nick-nikoliuk: That is the place where our sagas start working. We have no user action to start check is user signed in, so we start it here.

src/components/App/AppContainer.js
nick-nikoliuk: Not yet, which variant you prefer? I'll add correct rule.

src/components/LoginPage/LoginPageComponent.js
```diff
@@ -0,0 +1,72 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import slide1 from 'components/LoginPage/images/slide-1.png';
+import slide2 from 'components/LoginPage/images/slide-2.png';
+import slide3 from 'components/LoginPage/images/slide-3.png';
+import slide4 from 'components/LoginPage/images/slide-4.png';
+
+export class LoginPageComponent extends Component {
+  render() {
+    const authHandler = this.props.isGoogleApiInitialized ? this.props.requestAuthorization : null;
```
DarthVanger: I think this is the place where we should the spinner.

We're disabling an auth handler for the login buttons here.

We could hide the whole component instead and show text "Loading Google API...".

Showing spinner in sagas by applying a side DOM effect doesn't really look cool :)

We have the log of actions and can apply a middleware to show global messages for what is loading (actually we don't really need this messages anyways :).
Calling `showSpinner()` in each saga feels a bit redundant.

nick-nikoliuk: I agree with all notes except spinner. From my perspective spinner showing us process of saga. So we should use it in all sagas that should show to user loading status. It’s a async business logic and should be in saga, except we don’t need this spinner. 

For better code style and structure, we should use react spinner component. In this case in sagas will be presented only actions to show and hide spinner. Meanwhile, this actions will be saved status to store and react spinner will update his state. 

Am I understand correct?

src/components/App/AppSagas.js
```diff
@@ -0,0 +1,33 @@
...
+  console.info('Loading app...');
+
+  yield showSpinner('Loading Google Api');
+  yield googleApiLoader.load();
+  yield googleDriveApi.loadDriveApi();
+  yield thoughtStorage.scanDrive();
+  yield setPageAction(PAGES_ENUM.NOTES);
```
DarthVanger: yields look cool :) so better then promises.

Where are the errors are going to be handled though?

nick-nikoliuk: Sagas using try/catch construction for this. And as a result,  ACTION_NAME_FAILED will be dispatched. 
https://redux-saga.js.org/docs/basics/ErrorHandling.html

src/components/Spinner/SpinnerActions.js
nick-nikoliuk: removed

src/components/App/AppContainer.js
DarthVanger: Any variant will do :) Would be great to follow some popular style like airbnb.

I can see you have already added the rule to have one space and updated the code. Well done!

PR #27, opened 2019-06-24
 Implement App composite component p.3

nick-nikoliuk:
- move global css to globalStyles.js;
- replace thought to note;
- refactor NoteNameEditor component and helpers;


src/components/NoteNameEditor/NoteNameEditorComponent.js
```diff
@@ -0,0 +1,56 @@
...
+  render() {
+    const { onDeleteClick, onUploadFileClick } = this.props;
+
+    return (
+      <StyledNoteNameEditor>
+        <HeaderComponent
+          onDeleteClick={onDeleteClick}
+          onUploadFileClick={onUploadFileClick}
+        />
+        <StyledNoteEditorTextArea
```
DarthVanger: Are we now going to have every component prefixed with "Styled"? :) Whenever we need any styles.

nick-nikoliuk: Yes, It helps us to understand, that we are working with styled components, as we used with components parts like (*Component, *Container, etc.). This components are specific and sometime should receive some props to work properly with themes or styles overriding.

PR #33, opened 2019-07-30
WIP: refactor NotesMindMapView component;

nick-nikoliuk:
removed google api multiple initialization;
removed unnecessary comments;

### DIFF
--- a/src/components/NotesMindMapView/NotesMindMapViewContainer.js
+++ b/src/components/NotesMindMapView/NotesMindMapViewContainer.js
@@ -2,6 +2,7 @@ import { connect } from 'react-redux';
 import { action } from 'sagas';
 
 import {
+  CHANGE_NOTE_VIS_NETWORK_NOTE_ACTION,
   CHANGE_SELECTED_NOTE_ACTION,
   REQUEST_NOTE_TEXT_ACTION
 } from 'components/NotesMindMapView/NotesMindMapViewActions';
@@ -16,6 +17,7 @@ const mapStateToProps = ({ notesMindMap: { selectedNote } }) => {
 const mapDispatchToProps = () => ({
   requestNoteText: data => action(REQUEST_NOTE_TEXT_ACTION, data),
   changeSelectedNote: data => action(CHANGE_SELECTED_NOTE_ACTION, data),
+  changeVisNetworkNote: data => action(CHANGE_NOTE_VIS_NETWORK_NOTE_ACTION, data),
 });
 
 export const NotesMindMapViewContainer = connect(
--- a/src/components/NotesMindMapView/NotesMindMapViewSagas.js
+++ b/src/components/NotesMindMapView/NotesMindMapViewSagas.js
@@ -1,14 +1,18 @@
 import { all, call, put, takeEvery } from 'redux-saga/dist/redux-saga-effects-npm-proxy.cjs';
+import _ from 'underscore';
 
 import {
+  CHANGE_NOTE_VIS_NETWORK_NOTE_ACTION,
   changeNoteTextAction,
   changeSelectedNoteAction,
   REQUEST_NOTE_TEXT_ACTION
 } from 'components/NotesMindMapView/NotesMindMapViewActions';
 import noteStorage from 'storage/noteStorage';
 import { APP_INIT_SUCCESS } from 'components/App/AppActions';
+import siteGlobalLoadingBar from 'ui/spinner/site-global-loading-bar';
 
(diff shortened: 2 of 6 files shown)


src/components/NotesMindMapView/NotesMindMapViewContainer.js
```diff
@@ -16,6 +17,7 @@ const mapStateToProps = ({ notesMindMap: { selectedNote } }) => {
 const mapDispatchToProps = () => ({
   requestNoteText: data => action(REQUEST_NOTE_TEXT_ACTION, data),
   changeSelectedNote: data => action(CHANGE_SELECTED_NOTE_ACTION, data),
+  changeBrainVisNetworkNote: data => action(CHANGE_NOTE_BVN_ACTION, data),
```
DarthVanger: I don't understand what does it mean to change a brain vis network note.

Also "Brain" is the old name of the app :) It can be omitted now. LIke `changeVisNetworkNote`.

DarthVanger: And same for the event name - `CHANGE_NOTE_BVN_ACTION` B is for Brain which should be removed.

And I don't understand what this action means and does as well.

src/components/NotesMindMapView/NotesMindMapViewSagas.js
```diff
@@ -17,13 +21,67 @@ function* setRootNote() {
...
+
+function* fetchParentNote(noteId) {
+  const fetchingParentNote = spinner.create('loading parent note');
+  const parentNote = yield noteStorage.fetchParentNote(noteId);
+  fetchingParentNote.hide();
+  return parentNote;
+}
+
+function* saveNote(note) {
+  put(changeSelectedNoteAction(note));
```
DarthVanger: why do we need to change selected note to the one we're saving?

src/components/NotesMindMapView/NotesMindMapViewSagas.js
```diff
@@ -17,13 +21,67 @@ function* setRootNote() {
...
+  if (note.isNote) {
+    yield requestNoteText(note);
```
DarthVanger: why do we need to request note text when we're saving a note?

src/components/NotesMindMapView/NotesMindMapViewSagas.js
```diff
@@ -17,13 +21,67 @@ function* setRootNote() {
...
+  }
+}
+
+function* fetchChildNotes(note) {
+  const fetchingNotesSpinner = spinner.create('loading child notes');
+  fetchingNotesSpinner.show();
+  const childNote = yield noteStorage.fetchChildNotes(note);
+  fetchingNotesSpinner.hide();
```
DarthVanger: where is error handling? If `fetchChildNotes` fails, what happens?

src/components/NotesMindMapView/NotesMindMapViewSagas.js
```diff
@@ -17,13 +21,67 @@ function* setRootNote() {
...
+  return childNote;
+}
+
+function renderChildren(children, targetNote, brainVisNetwork) {
```
DarthVanger: should we have render functions in a separate file? So we can then easily change the rendering engine?

src/components/NotesMindMapView/NotesMindMapViewContainer.js
nick-nikoliuk: > I don't understand what does it mean to change a brain vis network note.

That action to say vis network that it should update note.
Probably will be simplified when all project become React.

nick-nikoliuk: > Also "Brain" is the old name of the app :) It can be omitted now. LIke `changeVisNetworkNote`.

fixed

src/components/NotesMindMapView/NotesMindMapViewSagas.js
nick-nikoliuk: Fixed, by removing that method. Name was not related

src/components/NotesMindMapView/NotesMindMapViewSagas.js
nick-nikoliuk: function had unrelated naming

src/components/NotesMindMapView/NotesMindMapViewSagas.js
nick-nikoliuk: done

PR #37, opened 2019-10-07
fix network tree loading and move create and delete note to sagas;

nick-nikoliuk:


src/components/NotesMindMapView/NotesMindMapViewSagas.js
```diff
@@ -79,10 +80,49 @@ function renderChildren(children, targetNote, visNetwork) {
   });
 }
 
+function createEmptyChild({ data: { parentId, visNetwork } }) {
+  const note = {
+    name: 'new2',
```
DarthVanger: why 'new2'? :)

PR #45, opened 2019-11-23
Fix #40

### LINKED ISSUE #40: Note name editor is not shown when a new note is created
When a new note is created, it should automatically enter the "edit" mode: show the note name editor at the very top of the screen. But it's not happening. Instead, there is an error in console.

edit-note

Related refactoring ticket: https://github.com/DarthVanger/NeuralNotes/issues/48

tomarcafe:
Passing component is no good, but not worse than passing visNetwork.
Anyway seems like vis will be changed in order to be able to move nodes

### DIFF
--- a/src/components/NotesMindMapView/NotesMindMapViewComponent.js
+++ b/src/components/NotesMindMapView/NotesMindMapViewComponent.js
@@ -105,7 +105,7 @@ export class NotesMindMapViewComponent extends Component {
   visNetworkDoubleClickHandler = event => {
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      this.props.createEmptyChild({ parentId: targetNoteId, visNetwork });
+      this.props.createEmptyChild({ parentId: targetNoteId, visNetwork, context: this });
     }
   };
 
--- a/src/components/NotesMindMapView/NotesMindMapViewSagas.js
+++ b/src/components/NotesMindMapView/NotesMindMapViewSagas.js
@@ -80,7 +80,7 @@ function renderChildren(children, targetNote, visNetwork) {
   });
 }
 
-function createEmptyChild({ data: { parentId, visNetwork } }) {
+function createEmptyChild({ data: { parentId, visNetwork, context } }) {
   const note = {
     name: 'new2',
     content: '',
@@ -105,7 +105,7 @@ function createEmptyChild({ data: { parentId, visNetwork } }) {
         parentNoteId: parentId,
       });
 
-      this.editNote(note.id);
+      context.editNote(note.id);
 
       return note;
     });


DarthVanger:
>Passing component is no good, but not worse than passing visNetwork.

I'm not sure where should the `visNetwork` stuff go, but the whole point of having sagas and reducers was to reduce the mess :) We should continue refactoring.

I think a correct approach would be:
- Declare `EDIT_NOTE` action
- In the saga, instead of calling `this.editNote(note.id)`, dispatch the `EDIT_NOTE` action
- Add a handler for the `EDIT_NOTE` action in the saga `takeEvery(EDIT_NOTE, editNote)`. Declare the `editNote()` function in the saga, move logic from the `editeNote()` function which we have in the component:
    ```
     editNote() {
        const note = noteStorage.findNoteById(targetNoteId);

        if (note.name === noteStorage.APP_FOLDER_NAME || !note.isNote) {
          return;
        }

        visNetwork.selectNote(targetNoteId);
     }
    ```
- In the `NotesMindMapViewReducer.js`, add a handler for `EDIT_NOTE` to change the state the same way as it is now changed in the component
    ```
     this.setState({
         currentViewedNote: note,
         selectedNote: {
          id: targetNoteId
       }
    });
    ```
- In the `NotesMindMapViewContainer.js` add `currentViewedNote` to the object returned by `mapStateToProps()`
- In `NotesMindMapViewComponent.js` change all occurrences of `this.state.currentViewedNote` to `this.props.currentViewedNote`

So basically we would split the `editNote()` function we currently have in the component into two pieces: one in saga, and one in reducer. And the component would only connect to Redux and react to state changes.

DarthVanger:
>Anyway seems like vis will be changed in order to be able to move nodes

What do you mean, changed in which way?

DarthVanger:
Do you mean that if we change `vis` to some react-compatible lib, we won't have to deal with this problem of passing `visNetwork` around? And so it doesn't make sense to be refactoring the component before we change `vis` to something `react-compatible`?

Maybe :)

Though maybe if we just update `vis` network in `componentDidUpdate()` based on the current state, it would be not too hard and not that bad?

I understand your fix is a quick dirty fix, and we're already passing `visNetwork` so why not do another hack and pass component context... But I don't want people to have their eyes bleeding when they look at the codebase :D

So I guess I'd prefer to keep the bug around for a while, and start refactoring ASAP, instead of merging this dirty fix.

tomarcafe:
> > Anyway seems like vis will be changed in order to be able to move nodes
> 
> What do you mean, changed in which way?

changed to another library - as "Change note parent functionality (move note)" seems tough via vis - it only drags nodes, no easy way to drag--change-parent. 
btw, there's a freelancer at ethlance willing to complete the project, i think you should check

tomarcafe:
> So basically we would split the `editNote()` function we currently have in the component into two pieces: one in saga, and one in reducer. And the component would only connect to Redux and react to state changes.

will try, interesting if i can or not lol

PR #70, opened 2020-01-11
Convert tree DS to graph 

geekyd:
Converts the tree data structure to vis-graph compatible graph.

Todo:
- [x] Remove redundant code
- [x] Fix coloring issue of nodes in the graph


src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -63,31 +62,43 @@ export class NotesMindMapComponent extends Component {
...
 
-    const targetNote = tree(rootNote).find(node => node.id === targetNoteId);
+    const targetNote = notes.find(note => note.id === targetNoteId);
 
     if (!targetNote) {
       throw new Error('noteClickHandler(): couldn\'t find targetNote: ', targetNoteId);
     }
 
-    this.props.changeSelectedNote(targetNote);
+    this.props.changeSelectedNote({ note: targetNote, edges: this.props.edges });
```
DarthVanger: Does `changeSelectedNote` requires info about `edges`?

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -63,31 +62,43 @@ export class NotesMindMapComponent extends Component {
...
   visNetworkClickHandler = event => {
-    const { rootNote, selectedNote } = this.props;
+    const { selectedNote } = this.props;
     const { isChangeParentModeActive } = this.props;
 
     this.props.onMindMapClick();
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      const targetNote = tree(rootNote).find(node => node.id === targetNoteId);
+      const targetNote = this.props.notes.filter(note => note.id === targetNoteId)[0]
```
DarthVanger: why using `filter()[0]` instead of `find()`?

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -97,33 +108,33 @@ export class NotesMindMapComponent extends Component {
         });
       } else {
         if (targetNote.id !== selectedNote.id) {
-          this.props.changeSelectedNote(targetNote);
+          this.props.changeSelectedNote({ note: targetNote, edges: this.props.edges });
```
DarthVanger: Does `changeSelectedNote` require `edges`?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -22,64 +21,73 @@ const defaultState = {
...
+          newState.edges.push({ from: child.parent.id, to: child.id });
+        });
       }
+
       return newState;
   };
 
   const handleEditNoteNameAction = () => {
-      const noteToEdit = data;
+      const noteToEdit = data.note ? data.note : data;
```
DarthVanger: that's weird

src/components/NotesMindMap/NotesMindMapSagas.js
```diff
@@ -41,8 +41,9 @@ function* requestNoteText(note) {
  * and redraw the network for new notes.
  */
 function* changeSelectedNote({ data }) {
-  const targetNote = data;
-  if (didNotAttemptToFetchChildren(targetNote)) {
+  const targetNote = data.note ? data.note: data;
```
DarthVanger: this is weird

src/storage/noteStorage.js
```diff
@@ -89,7 +89,7 @@ function remove(note) {
   return noteStorageApi
     .remove(note)
     .then(result => {
-      deleteNode(note);
+      //deleteNode(note);
```
DarthVanger: why?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -22,64 +21,73 @@ const defaultState = {
...
   const handleSelectedNoteChildrenFetchedAction = () => {
       const newState = cloneTreeInState();
       const childNotes = data;
+
       if (childNotes.length) {
-        newState.selectedNote.children = childNotes;
-      } else {
-        newState.selectedNote.hasNoChildren = true;
+        childNotes.forEach(child => {
+          newState.notes.push({ id: child.id, name: child.name, label: child.name, isNote: child.isNote /*group: (hasChildren ? 'parent' : 'children')*/ });
```
DarthVanger: Why not push the whole child? `newState.notes.push(child)` ?

DarthVanger: Previously it was
```
  newState.selectedNote.children = childNotes;
```
We just receive the children array and save it to the state. Without changing any properties for each node.
So why do we now need to change the properties for each node?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -22,64 +21,73 @@ const defaultState = {
   noteText: '',
   isChangeParentModeActive: false,
   rootNote: undefined,
+  notes: [],
```
DarthVanger: it's `noDes`, not `noTes` :)

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -22,64 +21,73 @@ const defaultState = {
...
-      child => child.id !== noteToDelete.id
-    );
+    let notes = []
+
+    newState.notes = newState.notes.filter(note => note.id !== noteToDelete.id)
+    notes = [ ...newState.notes ]
+    newState.edges = newState.edges.filter(edge => {
+      if (edge.from === noteToDelete.id) {
+        notes = notes.filter(note => note.id !== edge.to)
+      } else { return edge }
```
DarthVanger: What is going on here? :)

src/components/NotesMindMap/NotesMindMapComponent.js
geekyd: Passing to check if a note's children have been fetched.

src/components/NotesMindMap/NotesMindMapSagas.js
geekyd: Removed

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Removed

src/components/NotesMindMap/NotesMindMapContainer.js
```diff
@@ -13,12 +13,14 @@ import {
 } from 'components/NotesMindMap/NotesMindMapActions';
 import { NotesMindMapComponent } from 'components/NotesMindMap/NotesMindMapComponent';
 
-const mapStateToProps = ({ notesMindMap: { rootNote, selectedNote, noteText, showNoteNameEditor, isChangeParentModeActive } }) => {
+const mapStateToProps = ({ notesMindMap: { rootNote, selectedNote, noteText, showNoteNameEditor, isChangeParentModeActive, notes, edges } }) => {
```
DarthVanger: Do we still need `rootNote` here?

src/components/NotesMindMap/NotesMindMapContainer.js
```diff
@@ -13,12 +13,14 @@ import {
...
   return {
     selectedNote,
     showNoteNameEditor,
     noteText,
     rootNote,
```
DarthVanger: Do we still need `rootNote` here?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -99,9 +91,16 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
     return newState;
   }
 
+  const addRootToGraph = () => {
+    const newState = { ...state };
+    newState.rootNote = data;
+    newState.notes.push({ id: data.id, label: data.name });
```
DarthVanger: Isn't `push` mutating the state?

https://redux.js.org/recipes/structuring-reducers/immutable-update-patterns/#inserting-and-removing-items-in-arrays

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -99,9 +91,16 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
     return newState;
   }
 
+  const addRootToGraph = () => {
+    const newState = { ...state };
+    newState.rootNote = data;
```
DarthVanger: do we still need `rootNote`?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -140,6 +139,9 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
     const clonedState = { ...state };
```
DarthVanger: We don't have a tree anymore, so I guess we should remove the `cloneTreeInState` function?

One of the reasons for the refactoring was to get rid of this function and make reducers simpler.

src/components/NotesMindMap/NotesMindMapSagas.js
```diff
@@ -109,8 +109,8 @@ function* searchNoteSaga({ data }) {
   console.log('Search results: ', results);
 }
 
-function didNotAttemptToFetchChildren(note) {
-  return (!note.children || !note.children.length) && !note.hasNoChildren;
+function didNotAttemptToFetchChildren(note, edges) {
+  return !(edges ? edges.filter(edge => edge.from === note.id ).length > 0 : false);
```
DarthVanger: It looks like `edges` are initialized as `[]` in `selectRootNote`. `Boolean([])` is `true`.

So will `Boolean(edges)` ever be `false`, do we need to handle this situation?

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
+export function addGroupTagToNotes(notes, edges) {
```
DarthVanger: It's a bit weird that in the graph helper we work with `noTes`, and not with `noDes`.

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
...
+            note.group = edges.filter(edge => edge.from === note.id).length > 0 ? 'parent' : 'children' 
+            return note
+        })
+    }
+    return notes
+}
+
+export function removeNoteFromGraph(notes, edges, noteToDelete) {
+    notes = notes.filter(note => note.id !== noteToDelete.id) // remove the note from the array
+    edges = edges.filter(edge => {  // remove all the children of the deleted parent
```
DarthVanger: Why are you removing child nodes from the `edges` and not from the `nodes` ?

`edges` have no nodes, only connections.

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
...
+      if (edge.from === noteToDelete.id) {
+        notes = notes.filter(note => note.id !== edge.to)
+      } else { return edge }
```
DarthVanger: I think `filter()` expects a boolean to be returned. Why do you return an object?

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
...
+            return note
+        })
+    }
+    return notes
+}
+
+export function removeNoteFromGraph(notes, edges, noteToDelete) {
+    notes = notes.filter(note => note.id !== noteToDelete.id) // remove the note from the array
+    edges = edges.filter(edge => {  // remove all the children of the deleted parent
+      if (edge.from === noteToDelete.id) {
```
DarthVanger: I think you're deleting only the immediate children of a node (first level), omitting the nested children.

I.e. I don't see this function deleting children of children of a node.

It is not deleting the whole subtree, it is deleting only the first level of the subtree. If I understood the code correctly.

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
...
+        notes = notes.filter(note => note.id !== edge.to)
+      } else { return edge }
+
+      if (edge.to !== noteToDelete.id) return edge
```
DarthVanger: I think filter() expects a boolean to be returned. Why do you return an object?

src/helpers/graph.js
```diff
@@ -0,0 +1,22 @@
...
+    })
+
+    return { notes, edges }
+}
```
DarthVanger: Please always add EOL at the end of file.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -140,6 +139,9 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
     clonedState.rootNote = { ...state.rootNote };
```
DarthVanger: Do we still need `rootNote` ?

src/components/LoginPage/LoginPageComponent.js
```diff
@@ -91,7 +91,7 @@ export class LoginPageComponent extends Component {
             <H3><span>Neural</span>Notes is open source and free</H3>
             <GoogleLoginButton type="button" onClick={authHandler}>
               <img src={iconGoogle} />
-              <span>Sing up with Google</span>
+              <span>Sign up with Google</span>
```
DarthVanger: This is not related to the issue. It shouldn't be changed in the "Convert tree DS to graph" PR.

Otherwise if we merge this PR and then revert it, this typo will be committed again.

src/components/NotesMindMap/NotesMindMapComponent.js
geekyd: Done

src/components/NotesMindMap/NotesMindMapComponent.js
geekyd: Passing to check if a note's children have been fetched.

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: The reason for this is to store the directed graph as arrays of nodes & edges. The childNotes are pushed to the nodes array & the relationships are in edges. If we have childNotes inside the node then that defeats the whole purpose of the transition.

src/components/NotesMindMap/NotesMindMapContainer.js
geekyd: Removed

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: My bad :( Removed

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Removed

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Removed

src/components/LoginPage/LoginPageComponent.js
geekyd: reverted

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -16,7 +15,7 @@ export class NotesMindMapComponent extends Component {
       isChangeParentModeActive
     } = this.props;
 
-    const visGraph = this.treeToVisGraph();
```
DarthVanger: Why change `const` to `let` ?

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -16,7 +15,7 @@ export class NotesMindMapComponent extends Component {
...
+    let visGraph = { nodes: [...this.props.notes] , edges: [...this.props.edges] }
```
DarthVanger: Why copying the nodes and edges? Why not just
```
const { nodes, edges } = this.props;
```
?

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -69,61 +67,63 @@ export class NotesMindMapComponent extends Component {
...
-    const { rootNote, selectedNote } = this.props;
+    const { selectedNote } = this.props;
+    const { edges } = this.props;
     const { isChangeParentModeActive } = this.props;
 
     this.props.onMindMapClick();
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      const targetNote = tree(rootNote).find(node => node.id === targetNoteId);
+      const targetNote = this.props.notes.find(note => note.id === targetNoteId)
```
DarthVanger: Missing semicolon in the end of line

src/components/NotesMindMap/NotesMindMapReducer.js
DarthVanger: I mean why can't we change
```
 childNotes.forEach(child => {
          newState.notes.push({ id: child.id, name: child.name, label: child.name, isNote: child.isNote /*group: (hasChildren ? 'parent' : 'children')*/ });
```
To 
```
childNotes.forEach(child => newState.notes.push(child))
```
?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
-    targetNote.parent.children = targetNote.parent.children.filter(
-      child => child.id !== targetNote.id
-    );
+    notes = updateGroupOfOldParent(notes, edges, newParentId, oldParent.from)
     return {
-      ...newState,
+      ...state,
       isChangeParentModeActive: false,
       showNoteNameEditor: false,
+      edges, notes
```
DarthVanger: please keep 1 variable per line:
```
    return {
       ...state,
      isChangeParentModeActive: false,
      showNoteNameEditor: false,
      edges,
      notes,
   }
````

File: src/components/NotesMindMap/NotesMindMapReducer.js [hunk 28]
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
-    newState.selectedNote = noteToDelete.parent
-    noteToDelete.parent.children = noteToDelete.parent.children.filter(
-      child => child.id !== noteToDelete.id
-    );
-    return newState;
+    let notes = [ ...state.notes ]
+    let edges = [ ...state.edges ]
+    let graph = removeNoteFromGraph(notes, edges, noteToDelete)
+    notes = graph.notes
+    edges = graph.edges
```
DarthVanger: Looks a bit weird... Can't you make the `removeNoteFromGraph` function to be pure, so you don't have to clone the arrays before using it? So we could have
```
const { notes, edges } = removeNoteFromGraph(notes, edges, noteToDelete);
return { ...state, notes, edges };
```
Or even
```
return { ...state, ...removeNoteFromGraph(notes, edges, noteToDelete) };
```
?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
+      edges, notes
     };
   }
 
   const handleChangeSelectedNoteAction = () => {
-    const newState = cloneTreeInState();
-    newState.selectedNote = data;
-    return newState;
+    const newState = { ...state };
+    newState.selectedNote = data.note;
```
DarthVanger: `newState` is not used in the `return` statement.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
+    return { ...state, selectedNote: data.note };
+  }
+
+  const addRootToGraph = () => {
+    let notes = [ ...state.notes ]
+    notes.push({ id: data.id, label: data.name });
```
DarthVanger: In other places you push notes with more properties:
```
{ id: updatedNote.id, label: updatedNote.name, name: updatedNote.name, isNote: updatedNote.isNote }
```
Why does the root note has less properties?

src/components/NotesMindMap/NotesMindMapSagas.js
```diff
@@ -27,7 +27,7 @@ const LOADING_NOTE_MESSAGE = 'loading note contents...';
 let spinner = siteGlobalLoadingBar.create('mind map');
 
 function* selectRootNote({ data }) {
-  yield put(changeSelectedNoteAction(data));
+  yield put(changeSelectedNoteAction({ note: data, edges: [] }));
```
DarthVanger: Why does `changeSelectedNoteAction` require `edges` param?

Does selecting a note changes edges configuration?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+      return { ...node, group: 'children' }
+    } 
+    return node
+  })
+  return nodes
+}
+
+export function addGroupTagToNote(nodes, selectedNoteId) {
+  nodes = nodes.map(note => (note.isNote && note.id === selectedNoteId) ? { ...note, group: 'parent' } : note)
+  return nodes
```
DarthVanger: Why not semicolons in this file, if we have semicolons in all the existing files?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+      newEdges = newEdges.filter(e => e.to !== childNoteId)
+      return;
+    }
+  }
+}
+
+export function updateGroupOfOldParent(nodes, edges, newParentId, oldParentId) {
+  nodes = nodes.map(node => {
+    if (node.id === newParentId) return { ...node, group: 'parent' }
+    else if (node.id === oldParentId && edges.find(edge => edge.from === node.id) === undefined){
```
DarthVanger: No space before `)` and `{` in the end of this line

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
+export function removeNoteFromGraph(nodes, edges, noteToDelete) {
+  let newNodes = [ ...nodes ]
+  let newEdges = [ ...edges ]
+  let parentEdge = newEdges.find(edge => edge.to === noteToDelete.id)
+
+  removeChild(noteToDelete.id)
+  newNodes = newEdges.find(edge => edge.from === parentEdge.from) ? //check if there are siblings of deleted node if not revoke parant status
```
DarthVanger: 1. No space after `//`.
2. This line is too long (combined with the comment).
3. parAnt -> parEnt ?
4. A comment is a code smell. Can't we declare functions instead of using the comment?
    I.e.
   ```
   function doesDeletedNodeHasSiblings() { ... }
   function revokeParentStatus() { ... }
   ```
5. What does it mean to revoke a parent status? I don't remember the nodes to have any status. Which statuses can nodes have?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
+export function removeNoteFromGraph(nodes, edges, noteToDelete) {
+  let newNodes = [ ...nodes ]
+  let newEdges = [ ...edges ]
+  let parentEdge = newEdges.find(edge => edge.to === noteToDelete.id)
```
DarthVanger: From `parentEdge` we use only `parentEdge.from`.

Can't we just create a variable
```
const parentId = newEdges.find(edge => edge.to === noteToDelete.id).from;
```
to make the code more readable?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+  let newNodes = [ ...nodes ]
+  let newEdges = [ ...edges ]
+  let parentEdge = newEdges.find(edge => edge.to === noteToDelete.id)
+
+  removeChild(noteToDelete.id)
+  newNodes = newEdges.find(edge => edge.from === parentEdge.from) ? //check if there are siblings of deleted node if not revoke parant status
+             newNodes : newNodes.map(node => { return node.id === parentEdge.from ? { ...node, group: 'children' } : node })
+  return { notes: newNodes, edges: newEdges }
+
+  function removeChild(childNoteId) {
```
DarthVanger: When you call this function you pass `noteToDelete.id`. But the function expects `childNoteId`.

So is it `noteToDelete.id` or `childNoteId` ? A note to delete and a child note are a bit different things I think.

Also, in this file you operate with `nodes` but pass `noteToDelete` and `childNoteId`.
Let's use `noDe` word everywhere, since it's a graph utility and it knows nothing about `noTes`?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+    newNodes = newNodes.filter(note => note.id !== childNoteId)
+    if(newEdges.find(edge => edge.from === childNoteId)) { // check if the node has children
```
DarthVanger: A comment is a code smell. Extract into a function
```
const hasChildren = (node) => Boolean(newEdges.find(edge => edge.from === childNoteId));
```

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+      newEdges.forEach(edge => {
+        if (edge.from === childNoteId) {
+          removeChild(edge.to)
```
DarthVanger: I think `removeNode(edge.to)` would sound better.

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+          newEdges = newEdges.filter(e => (e.from !== childNoteId))
+        }
+      })
+      newEdges = newEdges.filter(e => e.to !== childNoteId) // remove the edge from parent to the deleted node
+    } else {
+      newEdges = newEdges.filter(e => e.to !== childNoteId)
```
DarthVanger: The same thing is done 2 lines above.

So you have
```
if (...) {
  newEdges = newEdges.filter(e => e.to !== childNoteId) // remove the edge from parent to the deleted node
} else {
   newEdges = newEdges.filter(e => e.to !== childNoteId)
}
```
- doing the same thing in `if` and `else`...

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+      return;
+    }
+  }
+}
+
+export function updateGroupOfOldParent(nodes, edges, newParentId, oldParentId) {
+  nodes = nodes.map(node => {
```
DarthVanger: 1. Changing a function param is not good...
2. Why not just `return node.map(...);`?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+    if (node.id === newParentId) return { ...node, group: 'parent' }
+    else if (node.id === oldParentId && edges.find(edge => edge.from === node.id) === undefined){
+      return { ...node, group: 'children' }
+    } 
+    return node
+  })
+  return nodes
```
DarthVanger: I think what you're doing in this function could be done easier and in less confusing way:
- if a node has children, it's group should be `parent`.
- otherwise, it's group should be `children`.

So instead of having a weird method named `updateGroupOfOldParent`, you could have a method 
```
const updateNodeGroup = node => hasChildren(node) ? {...node, group: 'parent'} : {...node, group: 'children'};
```

And then whenever you need to update any node's group, be it an old parent, or a new parent, you can just do
```
const oldParentWithUpdatedGroup = updateNodeGroup(oldParent);
```

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function addGroupTagToNote(nodes, selectedNoteId) {
```
DarthVanger: The name of the function sounds like it can be used to add a group tag to any note.

But it accepts a `selectedNoteId`. So is it supposed to be used only for selected note? In this case, should it b named `addGroupTagToSelectedNote` ?

src/helpers/graph.js
```diff
@@ -0,0 +1,42 @@
...
+  nodes = nodes.map(note => (note.isNote && note.id === selectedNoteId) ? { ...note, group: 'parent' } : note)
```
DarthVanger: Why does being in the `parent` group depends on whether the note is selected and on `note.isNote` ?

src/storage/noteStorage.js
```diff
@@ -58,7 +58,11 @@ function updateNoteName({ note, newName }) {
 }
 
 function remove(note) {
-  return noteStorageApi.remove(note);
+  return noteStorageApi
+    .remove(note)
+    .then(result => {
+      return result;
+    });
```
DarthVanger: Why this change?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
 
   const handleNoteNameUpdateRequestSuccessAction = () => {
     const updatedNote = data;
-    const newState = cloneTreeInState();
-    const noteToUpdate = tree(newState.rootNote).find(note => note.id === updatedNote.id);
-    noteToUpdate.name = updatedNote.name;
-    return newState;
+    let notes = [ ...state.notes ]
+    notes = notes.filter(note => note.id !== updatedNote.id)
+    notes.push({ id: updatedNote.id, label: updatedNote.name, name: updatedNote.name, isNote: updatedNote.isNote });
```
DarthVanger: `{ id: updatedNote.id, label: updatedNote.name, name: updatedNote.name, isNote: updatedNote.isNote }` is duplicated 2 times.

And also you have `{ id: newNote.id, name: newNote.name, label: newNote.name, isNote: newNote.isNote`, which I'd say duplicates the same thing for the 3rd time.

Might make sense to extract it into a function?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -15,31 +14,31 @@ import {
...
-  rootNote: undefined,
+  notes: [],
+  edges: []
 };
 
 export const notesMindMapReducer = (state = defaultState, { type, data }) => {
   const handleSelectedNoteChildrenFetchedAction = () => {
-      const newState = cloneTreeInState();
+      let notes = [ ...state.notes ]
+      let edges = [ ...state.edges ]
```
DarthVanger: Missing semicolons at the end of statements

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -49,59 +48,65 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
   const handleDeleteNoteRequestSuccessAction = () => {
     const noteToDelete = data;
-    const newState = cloneTreeInState();
-    newState.selectedNote = noteToDelete.parent
-    noteToDelete.parent.children = noteToDelete.parent.children.filter(
-      child => child.id !== noteToDelete.id
-    );
-    return newState;
+    let notes = [ ...state.notes ]
+    let edges = [ ...state.edges ]
```
DarthVanger: Missing semicolons at the end of statements

DarthVanger:
Merged manually as 1 commit: https://github.com/DarthVanger/NeuralNotes/commit/1240fbd17a3146871dad4b11a1a65fa90ef98c0b

Issue #72, opened 2020-01-15
Setup prettier for code formatting


DarthVanger:
- Setup prettier for code formatting.
- Add git pre-commit hook to validate the staged changes abort commit if it's not formatted according to prettier
- Run prettier to update the code formatting for all the files in the project, create a pull request.
    Ideally we would do it when there are no open pull requests, to not introduce conflicts.

PR #77, opened 2020-01-19
Issue #72. Setup Prettier

TomaTomoe:
Setup Prettier
Run Prettier for src/**/*.{js,css,html}


.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+    bracketSpacing: true,
+    jsxBracketSameLine: true,
+    singleQuote: true,
+    trailingComma: 'all',
+    // Override any other rules you want
+};
```
DarthVanger: Please add EOL and the end of file

PR #79, opened 2020-01-23
Removes lodash & underscore

### LINKED ISSUE #57: Remove `lodash.js` and `underscore.js`
Webpack bundle analyzer shows that `lodash.js` and `underscore.js` contribute significantly to the bundle size.

webpack-bundle-analyzer

We use only a few methods from these libs, which can be replaced with pure JS (see https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore) OR we can import only the function we use from lodash, e.g.
```
import isEqual from 'lodash.isequal';
```

Replace all the usages of underscore/lodash methods with pure JS OR import only the functions we use from lodash like `import isEqual from 'lodash.isequal'` (decide which one is better), and uninstall the libs.

geekyd:
Removes lodash & underscore dependencies.

fixes #57

### DIFF
--- a/src/helpers/debounce.js
+++ b/src/helpers/debounce.js
@@ -1,0 +1,14 @@
+export function debounce(func, wait) {
+  var timeout;
+
+  return function() {
+    var context = this,
+      args = arguments;
+
+    clearTimeout(timeout);
+
+    timeout = setTimeout(function() {
+      func.apply(context, args);
+    }, wait);
+  };
+}
--- a/package.json
+++ b/package.json
@@ -63,7 +63,6 @@
     "style-loader": "^0.23.1",
     "text-loader": "0.0.1",
     "uglifyjs-webpack-plugin": "^2.2.0",
-    "underscore": "^1.9.1",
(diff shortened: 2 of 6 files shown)


geekyd:
@DarthVanger Please have a look :)

src/helpers/debounce.js
```diff
@@ -0,0 +1,15 @@
...
+            args = arguments
+        
+        clearTimeout(timeout)
+
+        timeout = setTimeout(function() {
+            timeout = null;
+            func.apply(context, args);
+        }, wait)
+    }
+}
```
DarthVanger: Please add newline at end of file

src/helpers/debounce.js
```diff
@@ -0,0 +1,15 @@
...
+    var timeout;
+
+    return function() {
+        var context = this,
+            args = arguments
+        
+        clearTimeout(timeout)
+
+        timeout = setTimeout(function() {
+            timeout = null;
```
DarthVanger: What is the point of setting the timeout id to `null` ?

PR #80, opened 2020-01-26
Code clean-up after #70

### LINKED ISSUE #75: Address comments from PR "Convert tree DS to graph #70"
Address unresolved comments from a PR: https://github.com/DarthVanger/NeuralNotes/pull/70

dilpreetsio:
Code clean up after #70 

fixes #75

### DIFF
--- a/src/components/NotesMindMap/NotesMindMapComponent.js
+++ b/src/components/NotesMindMap/NotesMindMapComponent.js
@@ -13,12 +13,11 @@ export class NotesMindMapComponent extends Component {
       selectedNote,
       showNoteNameEditor,
       isChangeParentModeActive,
+      nodes,
+      edges,
     } = this.props;
 
-    let visGraph = {
-      nodes: [...this.props.notes],
-      edges: [...this.props.edges],
-    };
+    const visGraph = { nodes, edges };
 
     const visOptions = {
       interaction: {
@@ -71,9 +70,9 @@ export class NotesMindMapComponent extends Component {
     // if clicking on the current note, do nothing.
     if (targetNoteId === selectedNote.id) return;
 
-    const notes = this.props.notes;
+    const nodes = this.props.nodes;
 
-    const targetNote = notes.find(note => note.id === targetNoteId);
+    const targetNote = nodes.find(note => note.id === targetNoteId);
 
     if (!targetNote) {
       throw new Error(
@@ -96,7 +95,7 @@ export class NotesMindMapComponent extends Component {
     this.props.onMindMapClick();
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      const targetNote = this.props.notes.find(
+      const targetNote = this.props.nodes.find(
         note => note.id === targetNoteId,
       );
 
@@ -119,26 +118,26 @@ export class NotesMindMapComponent extends Component {
   };
 
   visNetworkDoubleClickHandler = event => {
-    const { notes } = this.props;
+    const { nodes } = this.props;
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      const targetNote = notes.find(node => node.id === targetNoteId);
+      const targetNote = nodes.find(node => node.id === targetNoteId);
       this.props.createEmptyChild({ parent: targetNote });
     }
   };
 
   visNetworkHoldHandler = event => {
-    const { notes } = this.props;
+    const { nodes } = this.props;
     if (VisNetworkHelper.clickedOnNote(event)) {
       let targetNoteId = VisNetworkHelper.getTargetNoteId(event);
-      const targetNote = notes.find(node => node.id === targetNoteId);
+      const targetNote = nodes.find(node => node.id === targetNoteId);
       this.editNote(targetNote);
     }
   };
 
   editNote(targetNote) {
-    const { notes } = this.props;
-    const note = notes.find(node => node.id === targetNote.id);
+    const { nodes } = this.props;
+    const note = nodes.find(node => node.id === targetNote.id);
 
     if (note.name === noteStorage.APP_FOLDER_NAME || !note.isNote) {
       return;
@@ -166,7 +165,7 @@ NotesMindMapComponent.propTypes = {
   isChangeParentModeActive: PropTypes.bool.isRequired,
   changeParentNote: PropTypes.func.isRequired,
   showNoteNameEditor: PropTypes.bool.isRequired,
-  notes: PropTypes.array.isRequired,
+  nodes: PropTypes.array.isRequired,
   edges: PropTypes.array.isRequired,
   onMindMapClick: PropTypes.func.isRequired,
   editNote: PropTypes.func.isRequired,
--- a/src/components/NotesMindMap/NotesMindMapReducer.js
+++ b/src/components/NotesMindMap/NotesMindMapReducer.js
@@ -14,8 +14,8 @@ import {
 import { ROOT_NOTE_FOUND_ACTION } from 'components/App/AppActions.js';
 import {
   updateGroupOfOldParent,
-  addGroupTagToNote,
-  removeNoteFromGraph,
+  addGroupTagToNodes,
+  removeNodeFromGraph,
 } from '../../helpers/graph';
 
 const defaultState = {
@@ -23,105 +23,119 @@ const defaultState = {
   showNoteNameEditor: false,
   noteText: '',
   isChangeParentModeActive: false,
-  notes: [],
+  nodes: [],
   edges: [],
 };
 
 export const notesMindMapReducer = (state = defaultState, { type, data }) => {
   const handleSelectedNoteChildrenFetchedAction = () => {
-    let notes = [...state.notes];
+    let nodes = [...state.nodes];
     let edges = [...state.edges];
 
     const childNotes = data;
     if (childNotes.length) {
       childNotes.forEach(child => {
-        notes.push({
-          id: child.id,
-          label: child.name,
-          name: child.name,
-          isNote: child.isNote,
-          group: 'children',
-        });
+        nodes = addNoteToGraph(nodes, child);
         edges.push({ from: child.parent.id, to: child.id });
       });
-      notes = addGroupTagToNote(notes, state.selectedNote.id);
+      nodes = addGroupTagToNodes(nodes, edges);
     }
-    return { ...state, notes, edges };
+    return {
+      ...state,
+      nodes,
+      edges,
+    };
   };
 
   const handleEditNoteNameAction = () => {
     const noteToEdit = data;
-    return { ...state, showNoteNameEditor: true, selectedNote: noteToEdit };
+    return {
+      ...state,
+      showNoteNameEditor: true,
+      selectedNote: noteToEdit,
+    };
   };
 
   const handleNoteNameUpdateRequestSuccessAction = () => {
     const updatedNote = data;
-    let notes = [...state.notes];
-    notes = notes.filter(note => note.id !== updatedNote.id);
-    notes.push({
-      id: updatedNote.id,
-      label: updatedNote.name,
-      name: updatedNote.name,
-      isNote: updatedNote.isNote,
-    });
-    return { ...state, notes };
+    let nodes = [...state.nodes];
+    nodes = nodes.filter(note => note.id !== updatedNote.id);
+    return {
+      ...state,
+      nodes: addNoteToGraph(nodes, updatedNote),
+    };
   };
 
   const handleCreateNoteSuccessAction = () => {
     const newNote = data;
-    let notes = [...state.notes];
+    let nodes = [...state.nodes];
     let edges = [...state.edges];
-    notes.push({
-      id: newNote.id,
-      name: newNote.name,
-      label: newNote.name,
-      isNote: newNote.isNote,
-      group: 'children',
-    });
     edges.push({ from: newNote.parent.id, to: newNote.id });
-    notes = addGroupTagToNote(notes, state.selectedNote.id);
-    return { ...state, notes, edges };
+    nodes = addGroupTagToNodes(addNoteToGraph(nodes, newNote), edges);
+    return {
+      ...state,
+      nodes,
+      edges,
+    };
   };
 
   const handleDeleteNoteRequestSuccessAction = () => {
     const noteToDelete = data;
-    let notes = [...state.notes];
-    let edges = [...state.edges];
-    let graph = removeNoteFromGraph(notes, edges, noteToDelete);
-    notes = graph.notes;
-    edges = graph.edges;
-    return { ...state, notes, edges };
+    const { nodes, edges } = removeNodeFromGraph(
+      state.nodes,
+      state.edges,
+      noteToDelete,
+    );
+    return {
+      ...state,
+      nodes,
+      edges,
+    };
   };
 
   const handleChangeParentNoteRequestSuccess = () => {
     const noteId = data.noteId;
     const newParentId = data.newParentId;
     let edges = [...state.edges];
-    let notes = [...state.notes];
+    let nodes = [...state.nodes];
     const oldParent = edges.find(edge => edge.to === noteId);
     edges = edges.filter(edge => edge.to !== noteId);
     edges.push({ from: newParentId, to: noteId });
 
-    notes = updateGroupOfOldParent(notes, edges, newParentId, oldParent.from);
+    nodes = updateGroupOfOldParent(nodes, edges, newParentId, oldParent.from);
     return {
       ...state,
       isChangeParentModeActive: false,
       showNoteNameEditor: false,
       edges,
-      notes,
+      nodes,
     };
   };
 
   const handleChangeSelectedNoteAction = () => {
-    const newState = { ...state };
-    newState.selectedNote = data.note;
-    return { ...state, selectedNote: data.note };
+    return {
+      ...state,
+      selectedNote: data.note,
+    };
   };
 
   const addRootToGraph = () => {
-    let notes = [...state.notes];
-    notes.push({ id: data.id, label: data.name });
-    return { ...state, notes };
+    let nodes = [...state.nodes];
+    nodes.push({ id: data.id, label: data.name, name: data.name });
+    return {
+      ...state,
+      nodes,
+    };
+  };
+
+  const addNoteToGraph = (nodes, newNote) => {
+    nodes.push({
+      id: newNote.id,
+      label: newNote.name,
+      name: newNote.name,
+      isNote: newNote.isNote,
+    });
+    return nodes;
   };
 
   switch (type) {
--- a/src/components/NotesMindMap/NotesMindMapReducer.js
+++ b/src/components/NotesMindMap/NotesMindMapReducer.js
@@ -69,12 +69,8 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
     let nodes = [...state.nodes];
     let edges = [...state.edges];
     edges.push({ from: newNote.parent.id, to: newNote.id });
-    nodes = addNodeToGraph(nodes, newNote);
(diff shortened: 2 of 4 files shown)


geekyd:
@DarthVanger Please have a look :)

src/helpers/graph.js
```diff
@@ -1,38 +1,46 @@
+function doesDeletedNodeHasSiblings(edges, parentId) {
+  return edges.find(edge => edge.from === parentId)
+}
+
+function revokeParentStatus(nodes, parentId) {
```
DarthVanger: What does it mean to revoke a parent status?

Do nodes have status?

DarthVanger: (Question from the following comment from point 5: https://github.com/DarthVanger/NeuralNotes/pull/70/files#r368251724)

src/helpers/graph.js
```diff
@@ -1,38 +1,46 @@
...
+  removeNote(noteToDelete.id);
+
+  newNodes = doesDeletedNodeHasSiblings(newEdges, parentId)
+    ? newNodes : revokeParentStatus(newNodes, parentId)
   return { notes: newNodes, edges: newEdges };
 
-  function removeChild(childNoteId) {
-    newNodes = newNodes.filter(note => note.id !== childNoteId);
-    if (newEdges.find(edge => edge.from === childNoteId)) {
+  function removeNote(noteId) {
```
DarthVanger: removeNo**T**e - > removeNo**D**e

src/components/NotesMindMap/NotesMindMapComponent.js
```diff
@@ -13,12 +13,11 @@ export class NotesMindMapComponent extends Component {
...
       isChangeParentModeActive,
+      notes,
+      edges
     } = this.props;
 
-    let visGraph = {
-      nodes: [...this.props.notes],
-      edges: [...this.props.edges],
-    };
+    const visGraph = { nodes: notes, edges };
```
DarthVanger: Would be great if it was
```
const visGraph = { nodes, edges };
```

src/helpers/graph.js
```diff
@@ -1,38 +1,46 @@
...
-
-  removeChild(noteToDelete.id);
-  newNodes = newEdges.find(edge => edge.from === parentEdge.from) //check if there are siblings of deleted node if not revoke parant status
-    ? newNodes
-    : newNodes.map(node => {
-        return node.id === parentEdge.from
-          ? { ...node, group: 'children' }
-          : node;
-      });
+  let parentId = newEdges.find(edge => edge.to === noteToDelete.id).from;
```
DarthVanger: Why not `const`? Is it changing below?

src/helpers/graph.js
```diff
@@ -1,38 +1,46 @@
...
+    ? newNodes : revokeParentStatus(newNodes, parentId)
   return { notes: newNodes, edges: newEdges };
 
-  function removeChild(childNoteId) {
-    newNodes = newNodes.filter(note => note.id !== childNoteId);
-    if (newEdges.find(edge => edge.from === childNoteId)) {
+  function removeNote(noteId) {
+    newNodes = newNodes.filter(note => note.id !== noteId);
+    if (hasChildren(newEdges, noteId)) {
       // check if the node has children
```
DarthVanger: useless comment. It says "if has children" below

src/helpers/graph.js
```diff
@@ -1,38 +1,46 @@
...
       newEdges.forEach(edge => {
-        if (edge.from === childNoteId) {
-          removeChild(edge.to);
-          newEdges = newEdges.filter(e => e.from !== childNoteId);
+        if (edge.from === noteId) {
```
DarthVanger: no **d** eId

src/helpers/graph.js
```diff
@@ -42,10 +50,9 @@ export function updateGroupOfOldParent(nodes, edges, newParentId, oldParentId) {
     }
     return node;
   });
-  return nodes;
 }
 
-export function addGroupTagToNote(nodes, selectedNoteId) {
+export function addGroupTagToNotes(nodes, selectedNoteId) {
```
DarthVanger: addGropuTagToNo **d** es

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -113,17 +112,31 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
...
-    return { ...state, notes };
+    notes.push({ id: data.id, label: data.name, name: data.name });
+    return { 
+      ...state, 
+      notes, 
+    };
   };
 
+  const addNoteToGraph = (notes, newNote) => {
+    notes.push({
```
DarthVanger: should this function be mutating the `notes` parameter?

DarthVanger:
There was a fix for `prettier` and `eslint` in `master`.
Please rebase.

src/helpers/graph.js
geekyd: Yes, we need to revoke the `parent` status of the node when we delete all its children to make sure the colors of nodes are rendered properly.

DarthVanger: I mean we have *node group*. There is no such thing as "parent status" in the code.

`revokeParentStatus` is not revoking the parent group. It is setting the group to `children`.

So the function name is confusing.

`revokeParentStatus` in fact sets the node group to `children`.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
-    let notes = [...state.notes];
-    notes = notes.filter(note => note.id !== updatedNote.id);
-    notes.push({
-      id: updatedNote.id,
-      label: updatedNote.name,
-      name: updatedNote.name,
-      isNote: updatedNote.isNote,
-    });
-    return { ...state, notes };
+    let nodes = [...state.nodes];
```
DarthVanger: If `addNoteToGraph` didn't mutate the `nodes`, we wouldn't have to clone them before calling that function.

Wouldn't that be more handy/intuitive?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
+    nodes = nodes.filter(note => note.id !== updatedNote.id);
+    return {
+      ...state,
+      nodes: addNoteToGraph(nodes, updatedNote),
```
DarthVanger: We have noDes, then addNoTe... then noDes and then again updatedNoTe.

Looks like a mess. This is an array of noDes, but some elements are noDes and others are noTes... Why?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
-      id: newNote.id,
-      name: newNote.name,
-      label: newNote.name,
-      isNote: newNote.isNote,
-      group: 'children',
-    });
     edges.push({ from: newNote.parent.id, to: newNote.id });
-    notes = addGroupTagToNote(notes, state.selectedNote.id);
-    return { ...state, notes, edges };
+    nodes = addGroupTagToNodes(addNoteToGraph(nodes, newNote), edges);
```
DarthVanger: When we create a note, its parent should either keep the "parent" group, or change the group from "children" to "parent", if it's the first child of the node.

So we need to check only 1 node: the parent node. We don't need to map all the nodes.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
-    let notes = [...state.notes];
-    let edges = [...state.edges];
-    let graph = removeNoteFromGraph(notes, edges, noteToDelete);
-    notes = graph.notes;
-    edges = graph.edges;
-    return { ...state, notes, edges };
+    const { nodes, edges } = removeNodeFromGraph(
+      state.nodes,
+      state.edges,
+      noteToDelete,
```
DarthVanger: passing `noTe` to `removeNoDe`

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
     const newParentId = data.newParentId;
     let edges = [...state.edges];
-    let notes = [...state.notes];
+    let nodes = [...state.nodes];
     const oldParent = edges.find(edge => edge.to === noteId);
     edges = edges.filter(edge => edge.to !== noteId);
     edges.push({ from: newParentId, to: noteId });
 
-    notes = updateGroupOfOldParent(notes, edges, newParentId, oldParent.from);
+    nodes = updateGroupOfOldParent(nodes, edges, newParentId, oldParent.from);
```
DarthVanger: If a node has children, its group should be "parent", otherwise its group should be "children".

Why do we care if this node is an old parent? We have the `oldParent` node. Just use its id to see in `edges` if it has any children or not. Why in addition to the `oldParent.id` we need to pass the `newParentId` ?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
+      selectedNote: data.note,
+    };
   };
 
   const addRootToGraph = () => {
-    let notes = [...state.notes];
-    notes.push({ id: data.id, label: data.name });
-    return { ...state, notes };
+    let nodes = [...state.nodes];
+    nodes.push({ id: data.id, label: data.name, name: data.name });
```
DarthVanger: Can we use `addNoteToGraph()` here?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -14,114 +14,128 @@ import {
...
+    };
+  };
+
+  const addNoteToGraph = (nodes, newNote) => {
+    nodes.push({
+      id: newNote.id,
+      label: newNote.name,
+      name: newNote.name,
+      isNote: newNote.isNote,
+    });
```
DarthVanger: So basically it's `nodes.push({ ...newNote, label: newNote.name })`?

src/helpers/graph.js
```diff
@@ -1,38 +1,45 @@
-export function removeNoteFromGraph(nodes, edges, noteToDelete) {
+function doesDeletedNodeHasSiblings(edges, parentId) {
```
DarthVanger: Why *deleted* node? Doesn't it work for any node?

Also, why can't we do it like `doesNodeHasSiblings(edges, nodeId)` ? Why do we have to pass the parentId? I think it's easy to find a parent knowing the nodeId?

src/helpers/graph.js
```diff
@@ -1,38 +1,45 @@
-export function removeNoteFromGraph(nodes, edges, noteToDelete) {
+function doesDeletedNodeHasSiblings(edges, parentId) {
+  return edges.find(edge => edge.from === parentId);
+}
+
+function revokeParentStatus(nodes, parentId) {
```
DarthVanger: For `doesDeletedNodeHasSiblings` one has to pass `edges` as the first parameter.

For `revokeParentStatus` one has to pass `nodes` as the first argument.

I think it might be confusing when using these functions.

Maybe would be more handy to make it like `doesDeletedNodeHasSiblings({ edges, parentId })` and `revokeParentStatus({ nodes, parentId })`?

DarthVanger: Also, why does it accept `parentId` ? Is it revoking parent status from the passed parent, or from a child of this parent?

I think it would be less confusing if it was `revokeParentStatus({ nodes, nodeId })`. So we understand that the parent status is revoked from the passed node.

src/helpers/graph.js
```diff
@@ -1,38 +1,45 @@
...
-      });
-  return { notes: newNodes, edges: newEdges };
+    : revokeParentStatus(newNodes, parentId);
+  return { nodes: newNodes, edges: newEdges };
 
-  function removeChild(childNoteId) {
-    newNodes = newNodes.filter(note => note.id !== childNoteId);
-    if (newEdges.find(edge => edge.from === childNoteId)) {
-      // check if the node has children
+  function removenode(nodeId) {
```
DarthVanger: 1. removeNode (camel case)
2. We have a function named `removeNodeFromGraph` and a function named `removenode`. What is the difference between the two? Does `removenode` removes node not from graph but from somewhere else? I believe it's confusing.

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: In the next line, we are filtering the nodes array.

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Currently, the functions & actions are using `note` instead of `node`, Do you want me to change everything to `note`?

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Do we need to add other attributes of the fetched data i.e `mimeType, parents, parent`?
If yes then we can go for it.

DarthVanger: Oh didnt think about those :)
No we dont need them, so you are doing the right thing.

src/components/NotesMindMap/NotesMindMapReducer.js
DarthVanger: Youre right. Guess I commented the wrong line of code.

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Done

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Done

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Currently, the functions & actions are using note instead of node, Do you want me to change everything to note?

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Done

src/helpers/graph.js
geekyd: Done

src/helpers/graph.js
geekyd: Done

src/helpers/graph.js
geekyd: Done

src/components/NotesMindMap/NotesMindMapReducer.js
DarthVanger: Since the `newNote` has other attributes like `mimeType`, so it's not a note? But a folder?
So this function should be accepting `noteFolder` instead of `newNote`?
And so this function does 2 things:
- creates a node object from a folder object
- adds the created node to the graph

The function name `addNodeToGraph` makes me think that it simply adds a node that is passed to the graph. But turns out it accepts a folder,  creates a new object from it, and adds that to the graph.
So it's a bit confusing. Can we do something to make things more clear?

Right now here `newNote` has `mimeType` property, which is a bit confusing, since notes have no `mimeType`. Google Drive files have `mimeType`.

src/components/NotesMindMap/NotesMindMapReducer.js
DarthVanger: I want no mess in the code :) If `node` and `note` is the same thing, let's rename everything to `node` (I think `node` would be better than `note`, because graphHelper should work with nodes, and the component should render nodes using vis-network).
If `node` and `note` are two different things, then we shouldn't be passing `nodes` to `addNoteToGraph()` function.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -12,116 +12,139 @@ import {
...
   const handleChangeParentNoteRequestSuccess = () => {
     const noteId = data.noteId;
     const newParentId = data.newParentId;
     let edges = [...state.edges];
-    let notes = [...state.notes];
-    const oldParent = edges.find(edge => edge.to === noteId);
+    let nodes = [...state.nodes];
+    const oldParentId = edges.find(edge => edge.to === noteId).from;
     edges = edges.filter(edge => edge.to !== noteId);
+    const oldParentHasChildren = edges.find(edge => edge.from === oldParentId);
```
DarthVanger: Why not use a function from the graph helper? https://github.com/DarthVanger/NeuralNotes/pull/80/files#diff-8481f6c43a5e2d6f4ea40d7417e1d388R1

You are duplicating the logic by repeating it here.

src/helpers/graph.js
```diff
@@ -1,55 +1,47 @@
...
+  return edges.find(edge => edge.from === nodeId);
+}
+
+function revokeParentStatus({ nodes, nodeId }) {
+  return nodes.map(node => {
+    return node.id === nodeId ? { ...node, group: 'children' } : node;
+  });
+}
+
+const hasChildren = (edges, nodeId) =>
```
DarthVanger: `hasChildren` duplicates `doesNodeHasChildren`

src/helpers/graph.js
```diff
@@ -1,55 +1,47 @@
...
-export function addGroupTagToNote(nodes, selectedNoteId) {
-  nodes = nodes.map(note =>
-    note.isNote && note.id === selectedNoteId
-      ? { ...note, group: 'parent' }
-      : note,
+export function addGroupTagToNodes(nodes, edges) {
+  return nodes.map(node =>
+    node.isNote && hasChildren(edges, node.id)
+      ? { ...node, group: 'parent' }
+      : node,
```
DarthVanger: Shouldn't you add `group: 'children'` for notes without children?

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -12,116 +12,139 @@ import {
...
     edges.push({ from: newParentId, to: noteId });
-
-    notes = updateGroupOfOldParent(notes, edges, newParentId, oldParent.from);
+    nodes = nodes.map(node => {
+      if (node.id === newParentId && node.isNote)
+        return { ...node, group: 'parent' };
+      else if (node.id === oldParentId && node.isNote && !oldParentHasChildren)
+        return { ...node, group: 'children' };
```
DarthVanger: I don't understand what are you doing here and why.

You've got the `addGroupTagToNodes` function in the graph helper https://github.com/DarthVanger/NeuralNotes/pull/80/files#diff-8481f6c43a5e2d6f4ea40d7417e1d388R41

Why are you doing something similar but a bit different in this reducer?

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Good point. It is confusing. We should rename it `newNode` instead, as we are basically adding a node to the graph. A node can be either a folder or a text file.

src/helpers/graph.js
geekyd: done

src/helpers/graph.js
geekyd: My bad. Done

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -12,116 +12,139 @@ import {
...
+    let nodes = [...state.nodes];
+    addNodeToGraph(nodes, { ...data, isNote: false });
+    return {
+      ...state,
+      nodes,
+    };
+  };
+
+  const addNodeToGraph = (nodes, newNote) => {
+    nodes.push({
```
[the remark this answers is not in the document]
geekyd: Done

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Done

src/components/NotesMindMap/NotesMindMapReducer.js
geekyd: Done

PR #87, opened 2020-01-27
Attachments

dhammma:
**Added libraries**

1. reselect (to make pretty memoized selectors)
2. redux-actions (to make pretty actions and reducers)
3. dot-props-immutable (to use as helpers in reducers)
4. react-transition-group (the easiest way to make simple animation for excellent UX)
5. awesome-icons (to use in the app)
6. abortcontroller-polyfill (polyfill for IE and some mammoth browser to make cancellable uploads)

**External code**

https://github.com/tanaikech/ResumableUploadForGoogleDrive_js

I used it because didn't found anything better to make google drive resumable uploads.
It would be great to refactor and make this code part pretty in the future, but in my opinion, currently, it's not necessary.

**Google drive API restrictions**

Google Drive API supports several types of file uploads:

https://developers.google.com/drive/api/v3/manage-uploads#upload_types

My solution implements resumable uploads. I found, that for files upper then 5Mb Google Drive API requires this type of file upload.

In the future would be great to add simple file upload for files less than 5Mb.

**Changes description**

It's implemented as a separate module `Attachments`. The module provides the next components:

* Upload button - handles add new uploads
* Status button - handles basic upload status and provides an opportunity to go to the attachments page
* Attachments page - shows all uploading files

The logic in the module is implemented with `redux` staff. The module has own `actions`, `reducer` and `saga`.

In my implementation I followed the next practice:

- all small state details should be available to use in components/any part of code (it allows use them to make excellent UI and UX)
- the logic should be implemented only in redux-saga

**Features list**

1. Full-state file uploading
2. Multiple files uploading
3. Uploads are cancellable by user
4. Uploads are handling network/api errors
5. Uploads can be restarted by user
6. Mindmap can use attachments state to display computable pseudo notes before real upload happens

In my implementation, I used my brain and google drive uploads as an example

**What is missed**

1. We need to make the replacement of pseudo notes and edges with real after upload succeeds. I didn't do it because I don't familiar with the mindmap component sagas/actions and reducers. It would be better if it would be done by someone in a separated task.


DarthVanger:
>Features list
>1. Full-state file uploading

What is full-state file uploading?

jsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "baseUrl": "./",
+    "paths": {
+      "components/*": ["src/components/*"],
+      "libs/*": ["src/libs/*"],
+      "auth": ["src/auth"]
+    }
+  }
+}
```
DarthVanger: Please add EOL to the end of file.

DarthVanger:
>abortcontroller-polyfill (polyfill for IE and some mammoth browser to make cancellable uploads)

How mammoth are those browsers? How big is the polyfill size? I don't really want to support mammoth browsers :) How many percent of users use those browsers?

src/components/Attachments/AttachmentsActions.js
```diff
@@ -0,0 +1,49 @@
...
+);
+
+export const fileUploadingFailure = createAction(
+  'FILE_UPLOADING_FAILURE',
+  (file, uploadFolderId, error) => ({ file, uploadFolderId, error }),
+);
+
+export const retryFileUpload = createAction(
+  'RETRY_FILE_UPLOAD',
+  (file, uploadFolderId) => ({ file, uploadFolderId }),
```
DarthVanger: Why do we need `(file, uploadFolderId)` function? Why not just use `({ file, uploadFolderId })` everywhere?

DarthVanger: The same question for all actions in this file.

src/components/Attachments/AttachmentsActions.js
```diff
@@ -0,0 +1,49 @@
...
+  }),
+);
+
+export const fileUploadingProgressUpdated = createAction(
+  'FILE_UPLOADING_PROGRESSS_UPDATED',
+  (file, uploadFolderId, progress) => ({ file, uploadFolderId, progress }),
+);
+
+export const fileUploadingSuccess = createAction(
+  'FILE_UPLOADING_SUCCESS',
```
DarthVanger: All the actions we have in this project end with `_ACTION`.
Not sure why Kolia did it this way. Is it fine that we will now have also actions without `_ACTION` in the end? :)

DarthVanger:
I ran the branch locally.
When I select a file for upload, open the attachmenets page, it had "cancelled" status. If I click "retry", the status is again "cancelled". Though I never cancel it. Video: https://drive.google.com/open?id=13X_Z6cyCjq3_hy30lrw9pyVYobEN_vjM

src/components/Attachments/AttachmentsPageHeader.js
```diff
@@ -0,0 +1,47 @@
...
+  width: 30px;
+  height: 30px;
+  border: none;
+  outline: none;
+  background: none;
+`;
+
+const AttachmentsPageHeader = () => {
+  const dispatch = useDispatch();
+  const canClearList = useSelector(Selectors.canClearUploadList);
```
DarthVanger: The page is named "attachments" but selector "canClear *upload* list".
I think would make more sense to call it UploadsPage then.
But you have used Attachments word in so many places... A lot of things to rename T_T.

src/components/Attachments/AttachmentsActions.js
```diff
@@ -0,0 +1,49 @@
...
+  'FILE_UPLOADING_FAILURE',
+  (file, uploadFolderId, error) => ({ file, uploadFolderId, error }),
+);
+
+export const retryFileUpload = createAction(
+  'RETRY_FILE_UPLOAD',
+  (file, uploadFolderId) => ({ file, uploadFolderId }),
+);
+
+export const clearList = createAction('CLEAR_LIST');
```
DarthVanger: `CLEAR_LIST` - which list, list of what?

src/components/Attachments/AttachmentsReducer.js
```diff
@@ -0,0 +1,107 @@
...
+};
+
+const normalDefaultState = {
+  uploads: [],
+};
+
+const defaultState = debug ? debugDefaultState : normalDefaultState;
+
+function updateFile(updates) {
+  return (state, { payload: { file, uploadFolderId, ...payload } }) =>
```
DarthVanger: `{ payload: { file, uploadFolderId, ...payload } ` - payload is used 2 times. I think the 2nd time it's used it's not the full payload, but only a part of it? So should be named differently?

src/components/Attachments/AttachmentsReducer.js
```diff
@@ -0,0 +1,107 @@
...
+          uploadFolderId,
+          ...updates,
+        };
+      }),
+    );
+}
+
+export const attachmentsReducer = handleActions(
+  {
+    [Actions.addUploadingFiles]: (
```
DarthVanger: Not sure what does `addUploadingFiles` mean.
If files are "uploading" (i.e. being uploaded), then where do we add them? They are already being uploaded, so should be already in the uploads list?
Do you mean "start uploading files" ? Or "add files to upload list" ?

src/components/Attachments/AttachmentsReducer.js
```diff
@@ -0,0 +1,107 @@
...
+      state,
+      { payload: { files, uploadFolderId } },
+    ) =>
+      dotProp.set(state, 'uploads', uploads => [
+        ...Array.from(files).map(file => ({ file, uploadFolderId })),
+        ...uploads,
+      ]),
+    [Actions.fileUploadInitialized]: updateFile({ status: 'initialized' }),
+    [Actions.fileUploadGetLocation]: updateFile({ status: 'getLocation ' }),
```
DarthVanger: `'getLocation '` has a space in the end. So is going to fail for `if (status === 'getLocation')` ?
Maybe would make sense to use constants (a kind of ENUM) for all the status stings? :)

src/components/Attachments/AttachmentsSagas.js
```diff
@@ -0,0 +1,139 @@
...
+        type: 'done',
+        result: result.result,
+      });
+      emitter(END);
+    },
+  };
+  const handleResult = handlersMap[result.status];
+
+  if (!handleResult) {
+    // @note: we don't need really handle this exception
```
DarthVanger: So why are we handling it? I'm confused :)

src/components/Attachments/AttachmentsSagas.js
```diff
@@ -0,0 +1,139 @@
...
+  const channel = yield call(createUploadFileChannel, file, uploadFolderId);
+
+  try {
+    while (true) {
+      const event = yield take(channel);
+
+      yield put(mapChannelEventToAction(event, file, uploadFolderId));
+    }
+  } catch (error) {
+    console.log('atata error', error);
```
DarthVanger: ? :))

src/components/Attachments/AttachmentsSagas.js
```diff
@@ -0,0 +1,139 @@
...
+  } finally {
+    console.log('channel closed');
```
DarthVanger: Which channel? "File upload channel closed" would be more descriptive for the console log.

src/components/Attachments/AttachmentsSagas.js
DarthVanger: :))

src/components/Attachments/StatusButtonComponent.js
```diff
@@ -0,0 +1,89 @@
...
+ * @todo: we need to do this component more information
+ * it should represent uploading status state:
+ * - when success
+ * - when partial success
+ * - when has errors
+ */
+const StatusButtonComponent = () => {
+  const dispatch = useDispatch();
+  const isStatusButtonVisible = useSelector(Selectors.hasUploadingFiles);
+  const isUploadingProcessing = useSelector(Selectors.isUploadingProcessing);
```
DarthVanger: What is the difference between `hasUploadingFiles` and `isUploadingProcessing` ? For me both sound like there are files which are being uploaded (upload is in progress).

Also `isUploadingProcessing` is not a correct English :) and so is confusing. Do you mean `isUploadInProgress`?
`hasUploadingFiles` is also confusing. It sounds like "does [something] have files which are in progress of upload?". Because you say upload **ing** files. Not "uploaded". Not "filesToUpload". And so on. Do you mean `isUploadsListEmpty`?

src/components/NotesPage/NotesPageComponent.js
```diff
@@ -16,6 +18,8 @@ export class NotesPageComponent extends Component {
         <NotesContentEditorContainer />
         {!isHelpViewed && <ControlsHelpComponent onClose={closeHelp} />}
         <SearchPanelContainer />
+        <StatusButtonComponent />
```
DarthVanger: Status of what? Might make sense to name it `UploadsStatusButtonComponent`?

src/libs/google-drive-uploader.js
```diff
@@ -0,0 +1,268 @@
...
+ * and we didn't found cool library for this feature
+ *
+ * We found it:
+ * GitHub https://github.com/tanaikech/ResumableUploadForGoogleDrive_js
+ *
+ * We can use it as it and refactor in the future
+ */
+/**
+ * ResumableUploadToGoogleDrive for Javascript library
+ * GitHub  https://github.com/tanaikech/ResumableUploadForGoogleDrive_js<br>
```
DarthVanger: >v1.0.0 (October 11, 2019)
>Initial release.
Wow we're so lucky the lib exists, it has been released so recently :)

src/libs/google-drive-uploader.js
```diff
@@ -0,0 +1,268 @@
+/**
+ * Unfortunately, Google is not providing the client for uploading files
+ * and we didn't found cool library for this feature
+ *
+ * We found it:
```
DarthVanger: >we didn't found cool library for this feature
>We found it:

So did we find it (the cool library) or we didn't find it? :) I'm confused.

src/libs/google-drive-uploader.js
DarthVanger: Which version/commit did you copy here?
There are 5 commits in that github repo.

The file is different from https://github.com/tanaikech/ResumableUploadForGoogleDrive_js/blob/master/resumableupload_js.js

You have added `export` in the end, created the `context` variable.

If a new version with a bug fix comes out, say `v1.0.1`, how are we going to update this file?

We could fork the repo, make the changes, add `package.json` and `npm install` it [directly from github](https://stackoverflow.com/questions/17509669/how-to-install-an-npm-package-from-github-directly) (without publishing to `npm`).

This way we would have the changes you made in a github repo's history. Also the commit you forked from would be visible in github this way.

src/sagas.js
```diff
@@ -30,5 +44,6 @@ export function* rootSaga() {
     notesInit(),
     noteMindMapInit(),
     notesContentEditorInit(),
+    fork(attachmentsInit),
```
DarthVanger: Why are we doing all inits except this one without `fork()`? :)
What's the point of `fork`ing the init, if it's only doing 
```
 yield all([
    takeEvery(Actions.addUploadingFiles, addUploadingFilesSaga),
    takeEvery(Actions.retryFileUpload, retryFileUploadSaga),
  ]);
```
?
It's only initializing the listeners, it's not doing any job?
If we fork the initialization, does it mean that the user can click the upload button before the saga initialized, so the button won't have any effect? I.e. if `addUploadingFiles` action would be fired before saga has started listening to this event?

src/components/Attachments/AttachmentsSagas.js
```diff
@@ -0,0 +1,139 @@
...
+  const handlersMap = {
+    initialize: () => emitter({ type: 'initialized' }),
+    getLocation: () => emitter({ type: 'getLocation' }),
+    start: () => emitter({ type: 'started' }),
+    Uploading: () =>
+      emitter({
+        type: 'uploading',
+        progress: result.progressNumber.current / result.progressNumber.end,
+      }),
+    Done: () => {
```
DarthVanger: Why `Uploading` and `Done` start with an uppercase letter, while all the other handlers start with a lowercase letter?

src/components/Attachments/AttachmentsReducer.js
DarthVanger: And what does the `getLocation` status mean? How can status be `getLocation`? Maybe `gettingLocation`? Or `gotLocation` ? And which location are we talking about?
I see it's coming from the ResumableUpload lib... But could you please add a comment of what the hell is that, why are we using it? :)

DarthVanger:
>What is missed

>We need to make the replacement of pseudo notes and edges with real after upload succeeds. I didn't do it because I don't familiar with the mindmap component sagas/actions and reducers. It would be better if it would be done by someone in a separated task.

Could you please create this task? Because you are the one who can describe the details (e.g. after which action should we update the pseudo notes with real notes, after which actions should we remove those pseudo notes).

dhammma:
>> What is full-state file uploading?

when we keep all state things in the app state
and when we can track anything that is happening

in comparison with the opposite - when we just make an action without an opportunity to track what's happening in whole details

jsconfig.json
dhammma: ok

dhammma:
>> How mammoth are those browsers? How big is the polyfill size? I don't really want to support mammoth browsers :) How many percent of users use those browsers?

incorrect translation :D  I mean old browsers in fact

>> How big is the polyfill size?

few hundred lines of code

>> How many percent of users use those browsers?

I don't know how to check it.  There is can I use info about supported browsers:

https://caniuse.com/#feat=abortcontroller

Not sure, that I understand `can i use` info correctly.

image

As I see, this feature available in Chrome since December 2019. 

**Arguments to use this polyfill**

1. high development standard - try to make your app accessible for as many as possible devices and users
2. on MDN `AbortController` is marked as an experimental feature
3. we don't pay for it - polyfill is small and we don't need to make efforts to add it

src/components/Attachments/AttachmentsActions.js
dhammma: popular practice to design action creators as functions

for functions, arguments are a more natural way to pass data

you can invoke action creator in next way:

```
const action = retryFileUpload(file1, folder1);
```

instead of:
```
const action = retryFileUpload({
  file: file1,
  uploadFolderId: folder1
});
```

if you decide to change the structure of action payload you don't need to change all code where you invoke your action creator as a function

additionally, it's an encapsulation of logic of creating action payload

instead of duplication a lot of code where you are making an object to pass as an argument to the action creator you are making it only in one place

dhammma:
>> I ran the branch locally.
>> When I select a file for upload, open the attachments page, it has "cancelled" status. If I click "retry", the status is again "cancelled". Though I never cancel it.

will push changes with fix

src/components/Attachments/AttachmentsActions.js
dhammma: Didn't noticed it.

image

In my opinion, it's unnecessary to add `_ACTION` suffix for action types. We will have an uncomfortable developer experience with it - `SELECTED_NOTE_CHILDREN_FETCHED_ACTION` too long and hard to read. Imagine a lot of action types with 4 words in the name. Too high price for the incomprehensible goals.

I can guess that `_ACTION` was added to mark constants for action types. It's not necessary when we are using `redux-action` or another way for action creators.

In my opinion, we can keep everything as-is for now. And decide one way for further development. I suggest to use `createAction` / `createActions` functions for making action creators and not use `_ACTION` prefix.

src/components/Attachments/AttachmentsActions.js
dhammma: My bad, will fix it

btw, if the app structure will grow, we'll have a way to make modular redux actions in a next way:

image

I didn't do it because we don't have agreements about redux actions, reducers, sagas and selectors structure. And this is not a necessary right now

src/components/Attachments/AttachmentsReducer.js
dhammma: >> I see it's coming from the ResumableUpload lib...

right

constants, surely, are a good idea, you are right

but I thought to defer it for the future for next reason:

They are coming from `third-party` code which I would like to refactor and restructure, I didn't want to do it now to save development time reasons

I'll add constants with comments in next push

src/components/Attachments/AttachmentsSagas.js
dhammma: I'll remove it - it seems like redundant for us, it was missed debug info

src/components/Attachments/AttachmentsReducer.js
dhammma: will replace it to `restPayload`

src/components/Attachments/AttachmentsReducer.js
dhammma: >> Not sure what does addUploadingFiles mean.

image

IMHO, everything is transparent

>> Do you mean "start uploading files" ? Or "add files to upload list" ?

no, I mean add uploading files, in russian "добавить выгружаемые файлы"

so, if `ADD_UPLOADING_FILES` is not enough transparent I can change it to `ADD_FILES_TO_UPLOAD_LIST`. Probably, it's better for understanding

src/components/Attachments/AttachmentsSagas.js
dhammma: we need to make an exception for developer - if he makes mistake when modify the code

but in fact, there shouldn't be exceptions here and I mean that we shouldn't handle this exception in outer scope

src/components/Attachments/StatusButtonComponent.js
dhammma: I'll try to rename them better

src/libs/google-drive-uploader.js
dhammma: >> If a new version with a bug fix comes out

we don't need new version, we need restructure this code and keep it in our repo

>> how are we going to update this file

if imagine, that we need it - we can see on the diff in the new version and apply changes
but I have a huge doubt that this situation is possible

You can think that we are not using a library. Because it's a pseudo library. You can think that it's working code snippet from the internet. In fact, it's something simple between library and code snippet. The author should rework it to make it a library.

Why is this not a library in my opinion?

The author didn't take care of ways to integrate it into the project. It has not package.json with version, it can't be imported as a module. A only way to add it to the project - including it to html (as described in readme). In 2020.

src/libs/google-drive-uploader.js
dhammma: it's not cool library :) but it's something. and something is better than nothing :)

src/components/Attachments/AttachmentsSagas.js
dhammma: will handle it

src/components/NotesPage/NotesPageComponent.js
dhammma: will rename it

src/components/Attachments/AttachmentsSagas.js
dhammma: legacy from cool uploading library

DarthVanger:
just one question about the redux-actions. What do you mean when you say `make pretty actions and reducers`? :)

DarthVanger:
why do u need such helpers ? "(to use as helpers in reducers)"
looks weird to me

DarthVanger:
Was merged as part of https://github.com/DarthVanger/NeuralNotes/pull/88

PR #88, opened 2020-02-02
Uploads

dhammma:
**Changes to:**

https://github.com/DarthVanger/NeuralNotes/pull/87

**What's changed:**

1. removed third-party code for uploading files
2. renamed all stuff `attachments` to `uploads`
3. renamed actions and selectors
4. fixed bug with `cancelled` at end of file upload
5. improved progress update
6. refactoring work
7. added support of real resumable uploads - when user clicks retry - we resume upload
8. removed statuses


src/components/NotesMindMap/NotesMindMapSelectors.js
```diff
@@ -0,0 +1,17 @@
+import { createSelector } from 'reselect';
+
+const getMindMapState = state => state.notesMindMap;
+
+export const getSelectedNote = createSelector(
+  getMindMapState,
+  mindMapState => mindMapState.selectedNote,
+);
+
+export const getSelectedNoteId = createSelector(getSelectedNote, selectedNote =>
```
DarthVanger: Why do we have part of selectors in `NotesMindMapContainer` and part in `NotesMindMapSelectors`?

src/components/Uploads/UploadStatusButton.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+`;
+
+/**
+ * @todo: we need to do this component more information
+ * it should represent uploading status state:
+ * - when success
+ * - when partial success
+ * - when has errors
```
DarthVanger: I don't understand what this `todo` is about.
Could you explain better what has been done and what to do?
Maybe create an issue on github instead of a `@todo`?
When should we do this todo? If we just leave it like it is, what is going to be "wrong" with this component?

DarthVanger: I would like to have an issue, instead of a `@todo` in the code, so I can prioritize it for before MVP or after MVP.
Nobody is going to touch this `@todo` because there is no task "do all the todos". Everybody is doing only issues from github.

src/helpers/redux.js
```diff
@@ -0,0 +1,49 @@
...
+ * Creates a function which transforms arg list to objekt with followed keys
+ *
+ * Example:
+ *
+ * const changeParentPayloadCreator = withPayload('nodeId', 'nextParentId');
+ * const payload = changeParentPayloadCreator(1001, 2000);
+ * // payload = {
+ * //   nodeId: 1001,
+ * //   nextParentId: 2000,
+ * // }
```
DarthVanger: omg what a useless stuff... :)

src/helpers/redux.js
```diff
@@ -0,0 +1,49 @@
...
+export const updateState = (name, updater) => {
+  const reducer = (state, { payload }) => {
+    const nextValue =
+      typeof updater === 'function' ? updater(state[name], payload) : updater;
+    const nextState = dotProp.set(state, name, nextValue);
+
+    return nextState;
+  };
+
+  return reducer;
```
DarthVanger: And another useless stuff :)

Why do you need to invent your own Redux helpers?... Why not just write the code like everybody?
I believe these functions make the code harder to understand without bringing much useful stuff :)

src/selectors.js
```diff
@@ -0,0 +1,11 @@
+import * as NotesMindMapSelectors from './components/NotesMindMap/NotesMindMapSelectors';
+import * as UploadsSelectors from './components/Uploads/UploadsSelectors';
+
+/**
+ * Global selectors
+ *
+ * Note!!!
+ * Don't import it to local component selectors
```
DarthVanger: Where should we import them?
Why do we need them?
Looks quite weird.
Can we refactor to get rid of global selectors?
Which problem requires global selectors?

src/components/NotesMindMap/NotesMindMapSelectors.js
dhammma: There is concept of layers of selectors:

1. reducer lever selector - implements some logic on reducer level
2. global level selector - includes all reducer level selectors and can be supplemented with some reusable selectors which combines reducer level selectors
3. component level selector - uses global selectors to select props for concrete component

It allows you to better organize your selectors, avoid circular dependencies (when your selectors use each other) and provides the way to reuse different modules selectors without applying some more complex technologies such as dependency injection.

In this example:

`NotesMindMapSelectors` - contains reducer level selectors
`NotesMindMapContainer` - contains component level selectors

dhammma: More detailed article:

https://cmichel.io/redux-selectors-structure

src/components/Uploads/UploadStatusButton.js
dhammma: Make status button more informative:

https://github.com/DarthVanger/NeuralNotes/issues/90

Refresh note children when upload completed:

https://github.com/DarthVanger/NeuralNotes/issues/91

src/helpers/redux.js
dhammma: We already discussed it. If you provide me the freedom to leave it as it - I'll prefer to keep those helpers :)

DarthVanger: sure :) Redux docs recommend creating helpers like these in the following article: https://redux.js.org/recipes/structuring-reducers/refactoring-reducer-example/

src/selectors.js
dhammma: >> Can we refactor to get rid of global selectors?

Yes, we can, but anyway we need to follow some practice to make selectors from different modules available to use each other without a circular dependency issue and with control how selectors are using each other.

>> Why do we need them?

To avoid circular dependencies when different selectors uses each other and to control the way how different selectors uses each other

>> Which problem requires global selectors?

https://randycoulman.com/blog/2016/09/27/modular-reducers-and-selectors/
https://randycoulman.com/blog/2018/06/12/solving-circular-dependencies-in-modular-redux/

DarthVanger: Okay. Please add to the comment where should we import these selectors. The comment says where not to use them, but doesn't explain where is it ok to use these global selectors. Where should we import them? In other global selectors? In `NotesMindMapContainer`? By "local selectors" do you mean selectors declared in `NotesMindMapComponent` or which ones?
You can link to the articles in the comment as well.
Otherwise it's not clear where to use these selectors (for me :) ).

src/helpers/redux.js
dhammma: removed it :)

src/selectors.js
dhammma: changed the comment

PR #102, opened 2020-05-18
Remove bootstrap and other redundant libs

### LINKED ISSUE #49: Remove bootstrap and other redundant libs
Remove `bootstrap` dependency from the project. We don't need it, and probably don't use it. If we're using it anywhere, replace it with a few lines of css.

Also remove the following libraries:
- [bootstrap](https://github.com/DarthVanger/NeuralNotes/blob/master/package.json#L36)
- [bootstrap-social](https://github.com/DarthVanger/NeuralNotes/blob/master/package.json#L37)
- [font-awesome](https://github.com/DarthVanger/NeuralNotes/blob/ca5b41541f5a1e66ed572b47fdedf26413866aa6/package.json#L61)
-   ["@fortawesome/fontawesome-svg-core": "^1.2.26", "@fortawesome/free-solid-svg-icons": "^5.12.0", "@fortawesome/react-fontawesome": "^0.1.8"](https://github.com/DarthVanger/NeuralNotes/blob/ca5b41541f5a1e66ed572b47fdedf26413866aa6/package.json#L21)

TomaTomoe:
Remove bootstrap and other redundant libs.

- removed bootstrap and font-awesome dependencies;

- compensated styles for login page;

- deleted font-awesome components and replased them via symbols.

 Resolved #49


DarthVanger:
The 1st line of the commit message is too long

image

See the seven rules https://chris.beams.io/posts/git-commit/

You have violated the "Limit the subject line to 50 characters" rule

rules

src/globalStyles.js
```diff
@@ -20,4 +27,20 @@ export const GlobalStyle = createGlobalStyle`
...
+  figure {
+    margin: 0;
+  }
+
+  .h1, .h2, .h3, .h4, .h5, .h6, h1, h2, h3, h4, h5, h6 {
+    font-family: inherit;
+    font-weight: 500;
+    line-height: 1.1;
+    color: inherit;
+  }
```
DarthVanger, line 45: Are you sure we need this on all pages?
Isn't it used only on login page?

TomaTomoe: I need these styles in Global or I must add it in every Slide from LoginPage. What would you prefer?

DarthVanger: Why on every slide? :)
Why can't we add it for the whole login page, don't we have a `LoginPageComponent`?

DarthVanger: We surely don't want to duplicate this code 5 times (for every slide).
If it's not possible to add these styles for the whole login page, and we have to keep them here in global, please add a comment
```
/**
 * Styles for Login page
 */
... code which is used only for login page ...

TomaTomoe: Sorry, that was my fault. I was wrong. I can do this style working from `LoginPageComponent`

Issue #97, opened 2020-05-18
Add top bar with logo and buttons (just design, no functionality)


DarthVanger:
<img width="706" alt="1-topbar" src="https://user-images.githubusercontent.com/6686679/82253153-1021a180-9959-11ea-9c69-5c098d6c8289.png">

PR #103, opened 2020-05-24
Add top bar. Without functionality. Only design

TomaTomoe:
- Move NotesMindMap and NotesContentEditor lower

- Add svg icons for futher buttons in TopBar

- Create TopBarComponent, TopBarContainer, TopBarStyles

- Change import TopBarComponent into NotesPageComponent

Implement #97


src/components/TopBar/TopBarComponent.js
```diff
@@ -0,0 +1,30 @@
+import React, { Component } from 'react';
+import UploadIcon from 'components/TopBar/images/ic_uploads.svg';
+import SearchIcon from 'components/TopBar/images/ic_search.svg';
+import MenuIcon from 'components/TopBar/images/ic_menu.svg';
```
DarthVanger, line 4: what is `ic`? :)

TomaTomoe: Will 'icon-uploads', 'icon-search' and 'icon-menu' be suitable? ^__^

DarthVanger: Yep :) 👍

DarthVanger:
image

`Change import TopBarComponent into NotesPageComponent`
mm.. what? :)
Change what? What do you mean by "change import"?
Import what into what? Change what? :)

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  height: 57px;
+  z-index: 2;
+`;
+
+export const StyledTopBar = styled.div`
+  position: absolute;
+  top: 0;
+  right: 0;
+  width: 100%;
+  height: 56px;
```
DarthVanger, line 17: Duplicated code

image

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  right: 16px;
+  top: 16px;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+`;
+
+export const StyledIcon = styled.img`
+  position: relative;
```
DarthVanger, line 53: Why do we need `position: relative` here? :)

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  height: ${props => props.height || '24px'};
+`;
+
+export const StyledGradientLine = styled.div`
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 1px;
+  z-index: 4;
```
DarthVanger, line 64: In reality we don't have 4 layers in Z direction here.
In reality It's all one layer. This line is just a border for the top bar. It's on layer `0`.
Why do we need to specify `z-index` here at all?

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  align-items: center;
+  color: #3c78c8;
+`;
+
+export const StyledControls = styled.div`
+  position: absolute;
+  width: 99px;
+  height: 24px;
+  right: 16px;
+  top: 16px;
```
DarthVanger, line 45: `  height: 24px;` duplicates `StyledHeader` height.
` top: 16px;` duplicates `StyledHeader` top.

I think it should be one number

image

DarthVanger: What if you just set `padding-top: 24px` for the `StyledTopBar`?
And use a usual position instead of absolute for header and controls.
It can be a flex container or `display: inline-block` or `float: right`.

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,71 @@
...
+  width: ${props => props.width || '24px'};
+  height: ${props => props.height || '24px'};
+`;
+
+export const StyledGradientLine = styled.div`
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 100%;
+  height: 1px;
```
DarthVanger, line 63: If we're adding a `1px` line with `position: absolute` at the bottom of an element, should we also set `margin-bottom: 1px` ? To ensure that the element and the bottom doesn't get cut by `1px` by this line.

src/components/NotesContentEditor/NotesContentEditorStyles.js
```diff
@@ -3,7 +3,7 @@ import styled from 'styled-components';
 export const StyledSelectedNotesContent = styled.div`
   position: absolute;
   z-index: 1;
-  top: 0;
+  top: 156px;
```
DarthVanger, line 6: Where does this number come from? :)

I wouldn't say it looks right...

image

Guess it should be `56px`? :)

Let's create a variable `topBarHeight` ? :)

`56px` is also used in `NotesMindMapStyles.js` on line 6.

src/components/TopBar/TopBarComponent.js
```diff
@@ -0,0 +1,30 @@
...
+export class TopBarComponent extends Component {
+  render() {
+    return (
+      <StyledTopBarWrapper>
+        <StyledTopBar>
+          <StyledHeader>NeuralNotes</StyledHeader>
+          <StyledControls>
+            <StyledIcon src={UploadIcon} />
+            <StyledIcon src={SearchIcon} />
+            <StyledIcon src={MenuIcon} width="12px" height="20px" />
```
DarthVanger, line 23: Why does this icon have different dimensions than the rest?

src/components/NotesContentEditor/NotesContentEditorStyles.js
DarthVanger: And `56px` is also sued for `StyledTopBar` height in `TopBarStyles.js`, line 17 :)

src/components/TopBar/TopBarStyles.js
DarthVanger: What do you think is better / easier in this case:
- use `position: absolute` for the `NeuralNotes` and the icons (like we do now0
- or use `flex` or `display: inline` or `float: right` to implement this layout

DarthVanger: For some reason I got a feeling that using `position: absolute` is not the right way for this case.
But maybe just because I used to not use `absolute` 🤔 
Your approach doesn't look too bad :)

src/components/TopBar/TopBarComponent.js
TomaTomoe: It's not obvious, but, please look at these three icons
image
here -> [template](https://www.figma.com/file/NUWWByWR7Pfpq3h0v9oTMM/Neural-Notes-Design-Beta-1.0?node-id=218%3A83)
They have the same size of outer-box but the distance between this boxs is different. Since I'm using flexbox, I can edit the size of the elements but not their "between distance"

src/components/TopBar/TopBarStyles.js
TomaTomoe: I agree with you, so I'll change this to `display: flex`. Only that I ask for it's to allow make margin- or padding-left for logo)

src/components/NotesContentEditor/NotesContentEditorStyles.js
TomaTomoe: The NotesMindMap was moved more than on 56px, because I want to show contrast of the bottom line of TopBar. This is temporary styles.

TomaTomoe: I'll change that to `top: 65px`. Is it good? So we will have nicer look.

TomaTomoe: It's will be look like that
image

src/components/TopBar/TopBarStyles.js
TomaTomoe: I took into account the height of this line, when set StyledTopBarWrapper height.

StyledTopBarWrapper's height = StyledTopBar's height + StyledGradientLine's height
57px=56px+1px

src/components/TopBar/TopBarStyles.js
DarthVanger: >.Only that I ask for it's to allow make margin- or padding-left for logo)

What do you mean? :)

src/components/TopBar/TopBarStyles.js
DarthVanger: Oh... Well that's not really obvious :)
Ideally we'd have a variable like  `bottomLineHeight`.
Or at least a comment
```
// 57px = 56px top bar height + 1px bottom gradient line height
height: 57px;
```
With variables would look nicer :)
```
height: topBarHeight + topBarBottomLineHeight
```

src/components/NotesContentEditor/NotesContentEditorStyles.js
DarthVanger: Sure 👍

src/components/TopBar/TopBarComponent.js
DarthVanger: Can we make this icon to be the same size as others? :)
I think the code will look nicer this way. More symmetric :)

Could you try editing `svg`'s dimensions?
image

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,58 @@
...
+
+export const StyledHeader = styled.header`
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 19px;
+  padding-left: 53px;
+  line-height: 22px;
+  display: flex;
+  align-items: center;
```
DarthVanger, line 32: >The CSS align-items property sets the align-self value on all direct children as a group.

I don't see `StyledHeader` having any children.
It has only text inside
```
  <StyledHeader>NeuralNotes</StyledHeader>
```

So I believe `align-items: center` doesn't have any effect here.

Same for `display: flex`.

src/components/TopBar/TopBarComponent.js
TomaTomoe: I've already try it. It's not solve the problem. In this case we will get next view:
image

src/components/TopBar/TopBarStyles.js
TomaTomoe: e853ffe456

I mean this distance. It must be reserved for further "back-button" 

image

src/components/TopBar/TopBarStyles.js
TomaTomoe: It would be great to create file with variables, that are used more than once. But do you think, that 'topBarBottomLineHeight' must be there too?

src/components/TopBar/TopBarStyles.js
TomaTomoe: Got it)

src/components/TopBar/TopBarStyles.js
DarthVanger: Oh, okay sure padding is fine :)

src/components/TopBar/TopBarComponent.js
DarthVanger: What did you change? :)
What's wrong with how it looks like?

DarthVanger: We can merge it as is. Not a blocker.

src/components/TopBar/TopBarStyles.js
DarthVanger: Without `topBarBottomLineHeight` variable we will have `height: topBarHeight + 1`.
Looking at this one might think

image

We don't have to create a separate file for variables.
We need it only here for now. So just declare it in this file
```
// height of the gradient line at the very bottom of the TopBar
const bottomLineHeight = 1;

const StyledGradientLine = styled.div`
  height: ${bottomLineHeight}px;
`

File: src/components/TopBar/TopBarStyles.js [hunk 10]
```diff
@@ -0,0 +1,58 @@
...
+  justify-content: space-around;
+  align-items: center;
+`;
+
+export const StyledIcon = styled.img`
+  width: ${props => props.width || '24px'};
+  height: ${props => props.height || '24px'};
+`;
+
+export const StyledGradientLine = styled.div`
```
DarthVanger, line 49: `GradientLine` is a bit confusing. Where is it on design? :)
`BottomLine` would be a bit clearer I think... Though still sounds odd :)

src/components/TopBar/TopBarStyles.js
TomaTomoe: Ok) Sure, I'll do it)

src/components/TopBar/TopBarStyles.js
TomaTomoe: As you wish)

src/components/TopBar/TopBarComponent.js
TomaTomoe: Ok, as you wish. But it still looks not perfect. I am killed by the different distances between the icons. Perhaps, as you are killed by the not perfect code)

src/components/TopBar/TopBarStyles.js
TomaTomoe: So, do I need to create "topBarHeight" constant too?

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,59 @@
...
+  background-color: #2a2a2a;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+  box-shadow: 0px 4px 6px rgba(0, 0, 0, 0.12), 0px 2px 4px rgba(0, 0, 0, 0.14),
+    0px 1px 2px rgba(0, 0, 0, 0.2);
+`;
+
+export const StyledHeader = styled.header`
```
DarthVanger, line 27: What do you mean by "header"?
Isn't the whole top bar a "header" in some way? :)

src/components/TopBar/TopBarComponent.js
```diff
@@ -0,0 +1,30 @@
...
+  StyledTopBarWrapper,
+  BottomLine,
+} from 'components/TopBar/TopBarStyles';
+
+export class TopBarComponent extends Component {
+  render() {
+    return (
+      <StyledTopBarWrapper>
+        <StyledTopBar>
+          <StyledHeader>NeuralNotes</StyledHeader>
```
DarthVanger, line 19: I don't think "header" is the best name here.
Usually a header is something like the top bar :) but here it's just a text that is displayed in a column. It's not even a row :)
[In Bootstrap it's called "brand"](https://getbootstrap.com/docs/4.0/components/navbar/): 
image

Also it may be called "logo". Or "app name" :)

image

DarthVanger: `TopBar` is a header in fact, by html5 semantic markup :)

image

And our thing is called `Logo` there :)

https://www.semrush.com/blog/semantic-html5-guide/

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,59 @@
...
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 19px;
+  padding-left: 53px;
+  line-height: 22px;
+  color: #3c78c8;
+`;
+
+export const StyledControls = styled.div`
```
DarthVanger, line 37: [I think it should be `StyledNav = styled.nav` :)](https://www.semrush.com/blog/semantic-html5-guide/)

image

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,59 @@
+import styled from 'styled-components';
+
+const bottomLineHeight = 1;
+const topBarHeight = 56;
+
+export const StyledTopBarWrapper = styled.nav`
```
DarthVanger, line 6: [I think it should be `header`](https://www.semrush.com/blog/semantic-html5-guide/)

src/components/TopBar/TopBarStyles.js
```diff
@@ -0,0 +1,59 @@
...
+  position: absolute;
```
DarthVanger, line 7: Why do we need a wrapper? :)

Can't we just put the bottom line inside the TopBar?

```
<Header>
  <Logo />
  <Nav />
  <BottomLine />
</Header>
```

Which benefits this wrapper brings us?
Which problem does it solve?

src/components/TopBar/TopBarComponent.js
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import UploadIcon from 'components/TopBar/images/icon-uploads.svg';
+import SearchIcon from 'components/TopBar/images/icon-search.svg';
+import MenuIcon from 'components/TopBar/images/icon-menu.svg';
+import {
+  StyledHeader,
+  StyledLogo,
+  StyledControls,
+  StyledIcon,
+  BottomLine,
```
DarthVanger, line 10: Why is everything `Styled...` except `BottomLine`? :)

Issue #99, opened 2020-05-18
Update the "+" add note button design


DarthVanger:
<img width="673" alt="3-plus-button" src="https://user-images.githubusercontent.com/6686679/82253340-7c040a00-9959-11ea-8ffd-444c0ae5e0da.png">

PR #104, opened 2020-05-26
Update the add note button design

TomaTomoe:
- Add new icon for add note button;

- Create styled component - StyledIcon;

- Change add note button size and backgroun-color in purpose of new design;

- Remove unnecessary classes due to not using Bootstrap anymore.

Implement #99


src/components/Uploads/UploadButton.js
```diff
@@ -35,6 +36,12 @@ const StyledUploadButton = styled.button`
   }
 `;
 
+const StyledIcon = styled.img`
+  position: relative;
+  width: ${props => props.width || '24px'};
+  height: ${props => props.height || '24px'};
```
DarthVanger, line 42: Looks like `props.height` and `props.width` are never used?
Also I don't think `position: relative` does something here? Since it has no children with absolute position.

If it's not used, let's remove it.

TomaTomoe: Ok, I'll  fix it)

PR #112, opened 2020-07-16
Note details preview: implement new design and functionality of note name and note content

TomaTomoe:
- Implement new design of note name and note content - Issue #124 
- Implement note name and note content change when selected note changes - Issue #114 
- Implement functionality of note name editing and note content editing - Issue #125


DarthVanger:
>Fix NoteDetailsReducer

- what was the bug? What did you fix?

src/components/NoteDetails/NoteDetailsComponent.js
```diff
@@ -1,19 +1,44 @@
...
+import {
+  StyledNoteNameEditor,
+  StyledDetailsScreen,
+  StyledNoteContentEditor,
+} from 'components/NoteDetails/NoteDetailsStyles';
 
 export class NoteDetailsComponent extends Component {
-  handleChange(e) {
-    this.props.onNoteNameChange(e.target.value);
+  handleChangeNoteName(e) {
```
DarthVanger, line 9: Please follow the naming used in the [React docs](https://reactjs.org/docs/forms.html#handling-multiple-inputs)

image

`handleChangeNoteName` -> `handleNoteNameChange`

src/components/NoteDetails/NoteDetailsComponent.js
```diff
@@ -1,19 +1,44 @@
...
+    this.props.onNoteNameChange({
+      newNoteName: e.target.value,
+      note: this.props.selectedNote,
+    });
+  }
+
+  handleChangeNoteContent(e) {
```
DarthVanger, line 16: `handleChangeNoteContent` -> `handleNoteContentChange`

src/components/NoteDetails/NoteDetailsComponent.js
```diff
@@ -1,19 +1,44 @@
...
 
   render() {
     return (
-      <StyledNoteNameEditor
-        onChange={e => {
-          this.handleChange(e);
-        }}
-        value={this.props.noteName}
-      />
+      <StyledDetailsScreen>
```
DarthVanger, line 25: I think it should be `StyledNoteDetailsScreen`, since it contains `StyledNoteNameEditor`, `StyledNoteContentEditor`. So to be consistent in naming: either we omit `Note` in all the names, or we add `Note` to all the names.

I think it's better to have `Note` in the name, because "details screen" is a very general name: details of what? `NoteDetails` makes the component name much more descriptive: this is for note details.

src/components/NoteDetails/NoteDetailsContainer.js
```diff
@@ -16,8 +19,11 @@ const mapStateToProps = ({
 };
 
 const mapDispatchToProps = () => ({
-  onNoteNameChange: noteName => {
-    action(EDITOR_NOTE_NAME_CHANGED_ACTION, noteName);
+  onNoteNameChange: ({ note, newNoteName }) => {
+    action(EDITOR_NOTE_NAME_CHANGED_ACTION, { note, newNoteName });
+  },
+  onNoteContentChange: ({ note, noteText }) => {
+    action(EDITOR_NOTE_CONTENT_CHANGED_ACTION, { note, noteText });
```
DarthVanger, line 26: the function name is `onNoteContentChange`, but it accepts `noteText`, not `noteContent`. Please use one name to avoid confusion.

src/components/NoteDetails/NoteDetailsReducer.js
```diff
@@ -1,22 +1,39 @@
...
+    case EDITOR_NOTE_CONTENT_CHANGED_ACTION:
+      return {
+        ...state,
+        noteText: data.noteText,
+      };
+    case CHANGE_NOTE_TEXT_ACTION:
+      return {
+        ...state,
+        noteText: data,
+      };
```
DarthVanger, line 36: Looks like `EDITOR_NOTE_CONTENT_CHANGED_ACTION` and `CHANGE_NOTE_TEXT_ACTION` do the same thing: they both return `{...state, noteText}`.

Why do we need both events? What is the difference between them?

Please use one name everywhere: `note_content` or `note_text`, in order to avoid confusion.

src/components/NoteDetails/NoteDetailsSagas.js
```diff
@@ -0,0 +1,47 @@
...
+  const newNote = yield noteStorage.updateNoteName({
+    note,
+    newName: newNoteName,
+  });
+  yield put(noteNameUpdateRequestSuccessAction(newNote));
+}
+
+function* updateNoteContent({ data: { note, noteText } }) {
+  let newNote = { ...note, content: noteText };
+  let savingNoteContentSpinner = spinner.create('saving note');
```
DarthVanger, line 27: `'saving note'` message is misleading here, it should be `'saving note content'`.

Otherwise it sounds like it's saving the whole note (both name and content), while in fact it's only saving note content.

src/components/NoteDetails/NoteDetailsReducer.js
TomaTomoe: I need to think, can't answer right now.

TomaTomoe: > Looks like `EDITOR_NOTE_CONTENT_CHANGED_ACTION` and `CHANGE_NOTE_TEXT_ACTION` do the same thing: they both return `{...state, noteText}`.
> 
> Why do we need both events? What is the difference between them?

I don't know if it important, but `EDITOR_NOTE_CONTENT_CHANGED_ACTION` returns `{...state, noteText: data.noteText}` when  `CHANGE_NOTE_TEXT_ACTION` returns `{...state, noteText: data}`

TomaTomoe:
Does this PR cover the Issue #121 too?

src/components/NoteDetails/NoteDetailsReducer.js
DarthVanger: Okay I took a closer look :)

`CHANGE_NOTE_TEXT_ACTION` is coming from `NoteMindMap`. It is a badly named action.
When selected note changes, note text is requested in `NotesMindMapSagas.js`. When the note text is fetched from Google Drive, `CHANGE_NOTE_TEXT_ACTION` is dispatched.
The action name should be `note text fetched`.

So we need to keep this event, otherwise when the selected note changes, our editor won't update the text.

DarthVanger:
> Does this PR cover the Issue #121 too?

We don't show "Note Edit" in the header and "OK" button.

Actually we need to ask Dimon why do we have this "Note Edit" screen. If the user can edit note on the Note Details screen.
Maybe Dimon's idea was that in Note Details the text is not editable? But when the user clicks on the text, the app switches into the Edit mode.
I asked him to make no Edit mode, to let the user edit note when he's viewing it, like in google docs. So probably we don't need the Edit screen at all, I will ask Dimon and close the #121 task if we don't need the Edit Note screen.

src/components/NoteDetails/NoteDetailsReducer.js
TomaTomoe: So, do you want me to change name of this action to NOTE_TEXTFETCHED?

DarthVanger: No :) We can't do it in this PR. We should rename it later in another PR, to `NOTE_CONTENT_FETCHED`.

PR #146, opened 2020-10-08
Issue #12 Finish Search notes functionality

jpalacio0612:
- Changed the search by folders instead of files.
- Redraw the mind map, showing that note and its children.

Final_Search

### DIFF
--- a/src/components/NotesMindMap/NotesMindMapReducer.js
+++ b/src/components/NotesMindMap/NotesMindMapReducer.js
@@ -10,6 +10,7 @@ import {
   CHANGE_PARENT_BUTTON_CLICKED_ACTION,
   CHANGE_PARENT_REQUEST_SUCCESS_ACTION,
   CHANGE_PARENT_REQUEST_FAIL_ACTION,
+  SEARCH_RESULT_CLICKED,
 } from 'components/NotesMindMap/NotesMindMapActions';
 import { ROOT_NOTE_FOUND_ACTION } from 'components/App/AppActions.js';
 import { addGroupTagToNodes, removeNodeFromGraph } from '../../helpers/graph';
@@ -164,6 +165,12 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
       return handleChangeParentNoteRequestSuccess();
     case CHANGE_PARENT_REQUEST_FAIL_ACTION:
       return { ...state, isChangeParentModeActive: false };
+    case SEARCH_RESULT_CLICKED:
+      return {
+        nodes: [data.note],
+        edges: data.edges,
+        selectedNote: data.note,
+      };
     default:
       return state;
   }
--- a/src/components/NotesMindMap/NotesMindMapReducer.js
+++ b/src/components/NotesMindMap/NotesMindMapReducer.js
@@ -168,7 +168,7 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
     case SEARCH_RESULT_CLICKED:
       return {
         nodes: [data.note],
-        edges: data.edges,
+        edges: [],
         selectedNote: data.note,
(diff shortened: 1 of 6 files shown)


src/components/NotesMindMap/NotesMindMapSagas.js
```diff
@@ -58,6 +61,21 @@ function* changeSelectedNote({ data: { note, edges } }) {
   }
 }
 
+function* handleSearchResultClick({ data: { note, edges } }) {
+  const targetNote = note;
+  if (didNotAttemptToFetchChildren(targetNote, edges)) {
```
DarthVanger, line 66: I don't think we need this `if()`.
It's used for `fetchChildNotes()`, to prevent refetching child notes when user clicks on a note which already has children fetched and displayed on the mind map.
But for search, we don't have such a situation.

src/components/NotesMindMap/NotesMindMapReducer.js
```diff
@@ -164,6 +165,12 @@ export const notesMindMapReducer = (state = defaultState, { type, data }) => {
       return handleChangeParentNoteRequestSuccess();
     case CHANGE_PARENT_REQUEST_FAIL_ACTION:
       return { ...state, isChangeParentModeActive: false };
+    case SEARCH_RESULT_CLICKED:
+      return {
+        nodes: [data.note],
+        edges: data.edges,
```
DarthVanger, line 171: let's just set the edges to `[]` here, instead of passing `[]` from action.

Action is `SEARCH_RESULT_CLICKED` and it passes the clicked search result as payload. Reducer decides what to do with edges. The action just says what happened: which search result was clicked.

src/components/SearchPage/SearchPageComponent.js
```diff
@@ -18,6 +19,11 @@ export function SearchPageComponent() {
     1000,
   );
 
+  const handleClick = note => {
+    const data = { ...note, label: note.name };
+    dispatch(searchResultClickedAction({ note: data, edges: [] }));
```
DarthVanger, line 24: I don't think we should pass `edges` here. The action should just tell what happened: a search result was clicked. It shouldn't say what should happened with mind map edges. That's responsibility of a reducer.

src/components/SearchPage/SearchPageComponent.js
```diff
@@ -27,8 +33,11 @@ export function SearchPageComponent() {
       {results &&
         results.map(result => (
           <>
-            <h2>{result.name}</h2>
-            <h3>{result.mimeType}</h3>
+            <h2
+              style={{ 'border-top': '1px solid black' }}
+              onClick={() => handleClick(result)}>
```
DarthVanger, line 38: `result` is passed to the `handleClick`, but `handleClick` accepts `note`. That's confusing. I believe the `handleClick` should accept `searchResult`, not `note`.

src/components/SearchPage/SearchPageComponent.js
DarthVanger, line 24: we have `note` as a function parameter, then we have `data` which is something different, and then we pass `{ note: data` as the action payload. So it's like we have two different notes here: `note` and `data`. This is confusing.

DarthVanger:
Why does this PR includes a commit saying `Component of TopBar deleted`?

image

I don't see TopBar being deleted in the diff. And I don't think we want to delete TopBar.

I guess what you wanted to say is "Delete unused import from TopBar.js" ?

Please change that commit message, because `Component of TopBar deleted` sounds like you have removed TopBar from the app.

Also, the next time please specify the issue number in commits.
E.g. `Basic search functionality` should be
```
Finish search notes feature

resolve #12
```

DarthVanger (approved):

Issue #93, opened 2020-02-16
Show last viewed note on startup


DarthVanger:
When the user opens the app, he/she should see the last viewed note and its children (currently when the app is opened, it shows the root note ("NeuralNotes") always).

The last viewed noteId can be saved to localStorage for example.

jpalacio0612:
Hello Ivan.

I will take this issue.

DarthVanger:
> Hello Ivan.
> 
> I will take this issue.

Okay! :) Assigned the issue to use and moved to `In progress`

jpalacio0612:
Hi Ivan.

I finish the first part to resolve this issue. But I have a question.
At the moment, to be more precise, I am storing the id of the last note in the localstorage and when I logout I clean the localstorage to avoid any error in case another different user enters the same browser. Is this operation correct? or the localstorage should persist even after logging out. 
This code is in a new branch called lastViewedNote

![lastnote](https://user-images.githubusercontent.com/65427911/96075915-b8d51e00-0e71-11eb-8c47-8e0f252c4cc7.gif)

DarthVanger:
Hi Jonathan!
Hm good question :)
I think you're right we should clear it when user logs out.

DarthVanger:
The 2nd part of the task was moved to a separate issue: https://github.com/DarthVanger/NeuralNotes/issues/154

PR #158, opened 2020-10-18
Issue #93 Show last viewed note on startup

jpalacio0612:
- Storage of the last note seen in the local storage.

- Render of the last note seen when the session starts.


src/api/google-drive-api.js
```diff
@@ -214,3 +215,15 @@ function findNotesByName(name = '') {
     });
   });
 }
+
+function findNotesById(lastViewedNoteId) {
```
DarthVanger, line 219: Note ID is unique. How can we find multiple notes by ID?

Let's rename it to `findNoteById` (without `S`) ?

src/components/App/AppSagas.js
```diff
@@ -27,7 +27,14 @@ export function* loadApp() {
   yield showSpinner('Loading Google Api');
   yield googleApiLoader.load();
   yield googleDriveApi.loadDriveApi();
-  const rootNote = yield noteStorage.scanDrive();
+  let rootNote;
+  const lastViewedNoteId = localStorage.getItem('lastViewedNoteId');
+  if (lastViewedNoteId) {
+    rootNote = yield googleDriveApi.findNotesById(lastViewedNoteId);
```
DarthVanger, line 33: you call `findNotes...` but assign it not to an array of notes, but to one note: `rootNote`.
So should it be `findNote...` ? (without `S`)

DarthVanger: Also, it's not `rootNote`. It's last viewed note :) Root note is `NeuralNotes`.

src/components/NotesMindMap/NotesMindMapSagas.js
```diff
@@ -63,6 +63,7 @@ function* changeSelectedNote({ data: { note, edges } }) {
 
 function* handleSearchResultClick({ data: { note } }) {
   const targetNote = note;
+  localStorage.setItem('lastViewedNoteId', targetNote.id);
```
DarthVanger, line 66: Why are we doing this only for the search result click, and not for a click on a note?
So if user did no search, there would be no last viewed note?
If user just starts with the root note and then selects another note, it should also be set as the last viewed note, because the user viewed it.

I think in `handleSearchResultClick` we can simply call `changeSelectedNote()`, and in `changeSelectedNote` we would set the `lastViewedNoteId`.

src/api/google-drive-api.js
```diff
@@ -214,3 +215,15 @@ function findNotesByName(name = '') {
     });
   });
 }
+
+function findNoteById(lastViewedNoteId) {
```
DarthVanger, line 219: Why is it a function of `lastViewedNoteId`? :)

This is for any note, so I believe it should be `function findNoteById(noteId)`

jpalacio0612: Done!

DarthVanger (dismissed):

DarthVanger:
@kvoncode please fix merge conflicts

DarthVanger (approved):

PR #161, opened 2020-10-19
Move logout button

kvoncode:
Logout button moved to top bar


DarthVanger:
Please delete `yarn.lock`. We don't need it.

kvoncode:
Can you please add it for me? I will much appreciate it, as I think **yarn** is much faster and better than npm

DarthVanger:
> Can you please add it for me? I will much appreciate it, as I think **yarn** is much faster and better than npm

You can add it to `.gitignore` or to [~/.gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore).

We don't need `yarn.lock` in the repo, so I don't want to push it. Others can add new npm modules, without updating `yarn.lock`,  then you'll update in in another PR... It would be messy.

kvoncode:
> > Can you please add it for me? I will much appreciate it, as I think **yarn** is much faster and better than npm
> 
> You can add it to `.gitignore` or to [~/.gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore).
> 
> We don't need `yarn.lock` in the repo, so I don't want to push it. Others can add new npm modules, without updating `yarn.lock`, then you'll update it in another PR... It would be messy.
> 
> This file is just for you, so having it in the repo would be confusing: others might think we're using `yarn` in CI... But we don't.

Ok. Will add to gitignore

PR #174, opened 2020-10-31
delete: unnecessary gapi load()

kvoncode:
loading Google API happens in 2 files:
2-files

in app it's:
app

and in login:
login

The app works fine with only 1 call to `load()`
fine

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,15 +2,19 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import * as ReactDOM from 'react-dom';
 import 'abortcontroller-polyfill/dist/polyfill-patch-fetch';
-
 import 'vis-network/dist/vis-network.css';
 import 'react-toastify/dist/ReactToastify.css';
 import 'sagas';
 import { App } from 'components/App/App';
 import { store, history } from 'sagas';
 import { GlobalStyle } from 'globalStyles';
-
 import { ConnectedRouter } from 'connected-react-router';
+
+// The app flow is following:
+// 1) The global code is executed. So it you have for example console.log() at the root of file it will execute
+// 2) redux-saga is imported before React, so saga's run() function will execute, and all effects will run in parallel
+// 3) The code inside React components will execute
+// 4) Saga's effects finish executing
 
 ReactDOM.render(
   <Provider store={store}>
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,7 @@ import { GlobalStyle } from 'globalStyles';
 import { ConnectedRouter } from 'connected-react-router';
 
 // The app flow is following:
-// 1) The global code is executed. So it you have for example console.log() at the root of file it will execute
+// 1) The global code is executed. So if you have for example console.log() at the root of file it will execute
 // 2) redux-saga is imported before React, so saga's run() function will execute, and all effects will run in parallel
 // 3) The code inside React components will execute
 // 4) Saga's effects finish executing
--- a/src/index.js
+++ b/src/index.js
@@ -11,7 +11,7 @@ import { GlobalStyle } from 'globalStyles';
 import { ConnectedRouter } from 'connected-react-router';
 
 // The app flow is following:
-// 1) The global code is executed. So if you have for example console.log() at the root of file it will execute
+// 1) The global code is executed. So if you have first line in js file like console.log() it will be execute
 // 2) redux-saga is imported before React, so saga's run() function will execute, and all effects will run in parallel
 // 3) The code inside React components will execute
 // 4) Saga's effects finish executing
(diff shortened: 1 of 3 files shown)


src/index.js
```diff
@@ -2,16 +2,20 @@ import React from 'react';
...
 import 'react-toastify/dist/ReactToastify.css';
 import 'sagas';
 import { App } from 'components/App/App';
 import { store, history } from 'sagas';
 import { GlobalStyle } from 'globalStyles';
-
 import { ConnectedRouter } from 'connected-react-router';
 
+// The app flow is following:
+// 1) The global code is executed. So if you have for example console.log() at the root of file it will execute
```
DarthVanger, line 14: What do you mean by "global code"? You mean code in `index.js` is executed?

kvoncode: Code in files that is executed. So for example we have `file.js` and inside it first line is
```
console.log(...)
```
It will be executed immediately on js load

DarthVanger: aha, so here
>So if you have for example console.log() at the root of file it will execute

you mean "at the root of any file" ?
I thought you mean just this file - `index.js`.

By the way I think it's called `top` of file, not `root` :) Not sure there is such a concept as "root of file".

kvoncode: Yes. Totally agree, perhaphs should change explanation on the one that wrote here

DarthVanger:
The current logic is as follows:
## If user is not signed in
- `api/google-api-loader` starts loading google client script `https://apis.google.com/js/api.js` and google `'client:auth2'`, and exports a `load()` function which returns a `promise`.
- `LoginPageSagas.js` calls `loginInit()`, which calls `api/google-api-loader` to get the `promise` because if user will try to login before `gapi` was loaded, this attempt will result in an error.
  - on `AUTH_SUCESS_SUCCESS` action (after user clicked login and signed in successfully), `AppSagas.js` call `loadApp()`
    - `loadApp()` calls `api/google-api-loader`, which again returns the same `promise`, which is already resolved, it wouldn't try to load `gapi` for the second time.
   - `loadApp()` calls `loadDriveApi()` which calls `gapi.client.load('drive', 'v3')` to load the GDrive API, which is needed for the app after user has signed in.

## If user is signed in
- `api/google-api-loader` starts loading google client script `https://apis.google.com/js/api.js` and google `'client:auth2'`, and exports a `load()` function which returns a `promise`.
- `AppSagas()` executes the `if (auth.signedIn())` branch, calling `loadApp()`.
-  `loadApp()` calls `api/google-api-loader`, which returns the `promise`, so it can wait for google client to be loaded, before trying to load GDrive client.
-  `loadApp()` calls `loadDriveApi()` which calls `gapi.client.load('drive', 'v3')` to load the GDrive API, which is needed for the app to show the notes of the signed-in user.

So the trick is that `api/google-api-loader`'s `load()` method returns a promise, so we can call `load()` as many times as we want, google api would be loaded only 1 time anyway.

This allowed us to call `api/google-api-loader`'s `load()` method both in `LoginPageSagas()` and in `AppSagas()`, without loading `gapi` twice.
If user is not authorized, we show Login page, and load only `gapi` and `client:auth2`, **without loading google drive client**.
If user is authorized, we load `gapi`, `client:auth2`, and google drive client: `gapi.client.load('drive', 'v3')`.

With the changes proposed in this PR, `google drive client` will be loaded for Login page, though login page doesn't need this client.

## Before
image

image

## After
image

image

## Impact
With `Fast 3G` internet this request adds extra `600ms` to the website load time

image

It won't block the user interaction, but the user will see the spinner, indicating that the website is not yet fully loaded:
image

## Is it worth it
I wouldn't say this extra request on login page is really critical: it's only `6KB` and those extra `600ms` are not blocking the UI.
But the refactoring you're proposing here also doesn't simplify the code too much as well.
Also, with the current approach the code is scalable: if we'd need to load something else in addition to google drive api for the non-login pages, we can add it to `loadApp()` without touching the login page load flow.
With the approach proposed in this PR, anything we add to the `appInit()` is going to be loaded for login page as well, even it's not needed for the login page.

My opinion is that it's better to keep the flow as it is, for better UX. Moreover this way code has better scalability, and we have separate flow for login page and the rest of the app.
The proposed refactoring simplifies the code, but it brings some problems with it, which I think outweight the benefits of the simplicity that this PR brings.

What do you think?

I'd agree though that the current code is a bit confusing, maybe we could make it more readable and obvious, still keeping the 2 separate flows for login page and for the rest of the app.

DarthVanger (approved):

kvoncode:
I see. So Google Drive API init function should call gapi initialization inside itself, because it's direct dependency.
(There is no way GD API can initialize without `gapi` object). 

`gapi` init should always be called,
but GD API should be called only on auth sucess.

I will address this issues

Issue #157, opened 2020-10-18
Uploads: use material list without background


DarthVanger:
Currently on uploads screen we have some gradient background for list items.

@TomaTomoe has updated the [design](https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836), and now the list items are transparent (have no background) like in Material UI.

User Material UI list and list item for upload items. https://material-ui.com/components/lists/#lists

PR #177, opened 2020-10-31
Uploads: used material list

Batsvinova-Anastasiia:
#157


src/components/Uploads/UploadsList.js
```diff
@@ -1,25 +1,31 @@
...
 
-const StyledContainer = styled.div`
-  display: flex;
-  flex-direction: column;
-`;
+const useStyles = makeStyles(() => ({
+  list: {
+    width: '100%',
+    height: '100%',
+    paddingLeft: 16,
```
DarthVanger, line 12: `16px` ?

src/components/Uploads/UploadsList.js
```diff
@@ -1,25 +1,31 @@
...
+import { makeStyles } from '@material-ui/core/styles';
 
-const StyledContainer = styled.div`
-  display: flex;
-  flex-direction: column;
-`;
+const useStyles = makeStyles(() => ({
+  list: {
+    width: '100%',
+    height: '100%',
```
DarthVanger, line 11: why `100%` height?

src/components/Uploads/UploadsListItem.js
```diff
@@ -8,19 +8,17 @@ import { colors } from '../../colors';
...
-`;
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import Divider from '@material-ui/core/Divider';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles(() => ({
+  listItem: {
+    padding: 0,
+    height: 64,
```
DarthVanger, line 19: `'64px'` ?

src/components/Uploads/UploadsListItem.js
```diff
@@ -34,14 +32,8 @@ const StyledRightContainer = styled.div`
...
-`;
-
-const StyledFileName = styled.div`
-  color: ${colors.white87};
-`;
-
-const StyledStatus = styled.div`
-  color: ${colors.white60};
+  padding-right: 16px;
+  padding-left: 16px;
```
DarthVanger, line 36: Why do we have 16px left padding here and in for the `<List>` in src/components/Uploads/UploadsList.js ?
Isn't it enough to specify left padding in one place only?

DarthVanger: Maybe it makes sense, can't see the DOM structure right now because the button doesn't redirect to the uploads page.

src/components/Uploads/UploadsList.js
Batsvinova-Anastasiia: fixed ✅

src/components/Uploads/UploadsList.js
Batsvinova-Anastasiia: now it looks like this 
image

this it's looks like this, if you remove height 100%
image

src/components/Uploads/UploadsListItem.js
Batsvinova-Anastasiia: fixed ✅

src/components/Uploads/UploadsListItem.js
Batsvinova-Anastasiia: with padding
image

without padding
image

src/components/Uploads/UploadsList.js
DarthVanger: What if we have more upload items so that we need scroll? Will `height: 100%` allow scroll?
Maybe we need `min-height: 100%` ?

Also, why do we have to cover light gray background with dark gray background?
I mean why does the background of Uploads page is different than the default page background?
In the [design](https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836) almost all pages have the same background as Uploads: dark gray. I don't see any page with the light gray background. Maybe we can just set the dark gray as default background?
Why do we need this light gray background beneath the dark gray background?

src/components/Uploads/UploadsListItem.js
DarthVanger: ok :)

src/components/Uploads/UploadsList.js
Batsvinova-Anastasiia: > What if we have more upload items so that we need scroll? Will height: 100% allow scroll?
Maybe we need min-height: 100% ?

Agree, fixed it.

Replaced light gray to dark gray.

DarthVanger (approved):

PR #185, opened 2020-11-13
178 bottom bar

Batsvinova-Anastasiia:


src/components/BottomBar/BottomBar.js
```diff
@@ -42,9 +53,62 @@ export const BottomBar = () => {
...
+  const [anchorEl, setAnchorEl] = React.useState(null);
+  const open = Boolean(anchorEl);
+
+  const handleClick = event => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
+  };
```
DarthVanger, line 65: Let's move the code for menu to a separate file (separate component).
Right now it's confusing why does `BottomBar` have `open`, `anchorEl` and `setAnchorEl`, `handleClick` and `handleClose`.
If we keep this code in `BottomBar.js` it would be a bit more logical to name these variables `isMenuOpen`, `menuAnchorEl`, `setMenuAnchorEl`, `handleOpenMenuButtonClick`, `handleMenuClose`. Because it is all related to the `Menu`, not to `BottomBar` itself.

If we extract this code to a separate file: `BottomBarMenu.js`, then we can don't need to add the "menu" word to every variable name, and the code will be more readable and clean.

src/components/BottomBar/BottomBar.js
```diff
@@ -42,9 +53,62 @@ export const BottomBar = () => {
 
   const selectedNote = useSelector(state => state.notesMindMap.selectedNote);
 
+  const [anchorEl, setAnchorEl] = React.useState(null);
+  const open = Boolean(anchorEl);
+
+  const handleClick = event => {
```
DarthVanger, line 59: Handle which click? I think it should be something like `handleOpenMenuButtonClick`

DarthVanger: Also, I think usually when the menu button is clicked while the menu is open, the menu would close.
So maybe it should be `handleToggleMenuButtonClick`, and it should be toggling the menu like
```
handleToggleMenuButtonClick = event => {
  if (!isMenuOpen) {
    // if menu is not open, open it
    setAnchorEl(event.currentTarget);
  } else {
     // if menu is already open, and user clicks the button again, close the menu
     setAnchorEl(null);
}
```

src/components/BottomBar/BottomBar.js
```diff
@@ -42,9 +53,62 @@ export const BottomBar = () => {
...
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleClose = () => {
+    setAnchorEl(null);
```
DarthVanger, line 64: It's not obvious why ` setAnchorEl(null);` will close the menu.
I would prefer
```
const handleMenuClose = () => {
    setIsMenuOpen(false);
}
```

src/components/BottomBar/BottomBar.js
Batsvinova-Anastasiia: Thank you for the good comment. Totally agree with you. I moved the menu-related code to the separated file

src/components/BottomBar/BottomBar.js
Batsvinova-Anastasiia: > So maybe it should be handleToggleMenuButtonClick, and it should be toggling the menu like

Seems like we actually don't need it.  In fact, the second click on the icon button will close the menu. Menu closing is achieved in another way - mui menu renders overlay and if a user clicks anywhere outside menu - menu would be closed without changing openMenu/closeMenu to toggleMenu

src/components/BottomBar/BottomBar.js
Batsvinova-Anastasiia: Agree with you. But I achieved more transparency in another way

DarthVanger (approved):
image

PR #205, opened 2021-04-15
Fix attributes BottomBarMenu.js error.

Cray-code:
Fix error:

```react-dom.development.js:88 Warning: React does not recognize the `aperProps` prop on a DOM element.```


DarthVanger (approved):
Well done, thank you!

PR #208, opened 2021-04-16
Correct design NotesPageTopBar and LogoutButton.

Cray-code:
The component <ToolBar> (inside <NotesPageTopBar>) has inline styles with media queries, so there can be 3 component heights: 48px, 56px and 64px, depending on the window width and device orientation.

### DIFF
--- a/src/components/LogoutButton/LogoutButtonComponent.js
+++ b/src/components/LogoutButton/LogoutButtonComponent.js
@@ -1,14 +1,10 @@
 import React from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
-import { colors } from 'colors';
+import { Typography } from '@material-ui/core';
 
 const useStyles = makeStyles({
-  logoutButton: {
-    fontSize: '16px',
-    lineHeight: '24px',
-    color: colors.white87,
-    marginRight: '22px',
+  button: {
     textTransform: 'capitalize',
   },
 });
@@ -16,8 +12,8 @@ const useStyles = makeStyles({
 export function LogoutButtonComponent({ logout }) {
   const classes = useStyles();
   return (
-    <Button className={classes.logoutButton} onClick={logout}>
-      Logout
+    <Button className={classes.button} onClick={logout}>
+      <Typography variant="subtitle1">Logout</Typography>
     </Button>
   );
 }
--- a/src/components/LogoutButton/LogoutButtonComponent.js
+++ b/src/components/LogoutButton/LogoutButtonComponent.js
@@ -1,15 +1,23 @@
-import React, { Component } from 'react';
-
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
+import { colors } from 'colors';
 
-export class LogoutButtonComponent extends Component {
-  render() {
-    return (
-      <Button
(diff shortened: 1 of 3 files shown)


src/components/LogoutButton/LogoutButtonComponent.js
```diff
@@ -1,15 +1,23 @@
-import React, { Component } from 'react';
-
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
+import { colors } from 'colors';
+
+const useStyles = makeStyles({
+  logoutButton: {
```
DarthVanger, line 7: I believe these styles are not specific to the `logoutButton`.
All dropdown menus in the app have the same style:
image

image

image

So I think we should apply this style to the MUI `MenuItem` in the [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L81). We already have some styles for the MenuItem there:
```
    MuiMenuItem: {
      root: {
        background: colors.dialogsGray,
        minHeight: '32px',
      },
    },
```

src/components/NotesPage/NotesPageTopBar.js
```diff
@@ -36,6 +34,17 @@ export const NotesPageTopBar = () => {
...
+  const pageWidth = document.documentElement.scrollWidth;
+  const pageOrientation = window.orientation ? window.orientation : 0;
+  let anchorPositionMenu = { top: 58, left: 0 };
+  if (pageWidth < 600) {
+    if (pageOrientation) {
+      anchorPositionMenu = { top: 42, left: 0 };
+    } else {
+      anchorPositionMenu = { top: 50, left: 0 };
+    }
+  }
```
DarthVanger, line 46: Maybe a better solution would be to use `anchorEl` ? https://stackoverflow.com/a/52551100/1657101

And set `anchorOrigin={{ vertical: "bottom",`, so we will have the menu appearing always at the bottom of the hamburger menu, independent of the appbar height.

Cray-code: The material-ui developers closed this ticket in October, and there is still no change on the site https://material-ui.com/components/menus/.
I parsed the example https://material-ui.com/components/menus/#customized-menus but didn't understand the importance of getContentAnchorEl = {null}
But it doesn't work without it.

src/components/LogoutButton/LogoutButtonComponent.js
Cray-code: I used styles ```<Typography>``` from the theme, but in order to remove ```textTransform: 'capitalize'```, instead of a button, we need to use another tag, not a ```<button>```.

src/components/LogoutButton/LogoutButtonComponent.js
```diff
@@ -1,15 +1,19 @@
-import React, { Component } from 'react';
-
+import React from 'react';
+import { makeStyles } from '@material-ui/core/styles';
 import Button from '@material-ui/core/Button';
+import { Typography } from '@material-ui/core';
+
+const useStyles = makeStyles({
+  button: {
+    textTransform: 'capitalize',
```
DarthVanger, line 8: I don't think we need this style. We have Logout starting with a capital letter already in the code:
```
      <Typography variant="subtitle1">Logout</Typography>
```
So I believe we don't need to capitalize it. It's already capitazlied.

DarthVanger (approved):

Issue #182, opened 2020-11-08
Implement topbar for Note Details page


DarthVanger:
Implement topbar for Note Details page as per design:

https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836

![image](https://user-images.githubusercontent.com/6686679/98461561-ce5e0f00-21b5-11eb-9e8d-0295d218349f.png)

Currently the Note Details page has no topbar:

![image](https://user-images.githubusercontent.com/6686679/98461581-f188be80-21b5-11eb-885a-83ca687150e7.png)

## We have similar dropdown on the notes mind map page
We already have the dropdown in the bottom bar on the notes mind map page, maybe the code can be reused or copy/pasted:

![image](https://user-images.githubusercontent.com/6686679/98461724-4416aa80-21b7-11eb-96cb-ba61504216c0.png)

PR #210, opened 2021-04-16
Refactor styles NoteDetailsPage

Cray-code:
In the project uses connected-react-router, and for some reason <Link> is used for navigation, although {push} should be used [connected-react-router](https://github.com/supasate/connected-react-router/blob/master/FAQ.md#how-to-navigate-with-redux-action)
Perhaps the component should use <TopBar> instead of <BackButtonWrapper> as a template for the topbar.
Or is it a separate ticket?

### DIFF
--- a/src/components/NoteDetails/NoteDetailsPage.js
+++ b/src/components/NoteDetails/NoteDetailsPage.js
@@ -4,16 +4,16 @@ import IconButton from '@material-ui/core/IconButton';
 import ArrowBackIcon from '@material-ui/icons/ArrowBack';
 import { Link } from 'react-router-dom';
 import styled from 'styled-components';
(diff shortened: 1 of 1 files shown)


DarthVanger:
> In the project uses connected-react-router, and for some reason is used for navigation, although {push} should be used [connected-react-router](https://github.com/supasate/connected-react-router/blob/master/FAQ.md#how-to-navigate-with-redux-action)
> Perhaps the component should use instead of as a template for the topbar.
> Or is it a separate ticket?

Yes let's do it in a separate pull request.

We do have a separate ticket for the NotesPage topbar: https://github.com/DarthVanger/NeuralNotes/issues/182

DarthVanger (approved):

PR #211, opened 2021-04-16
Refactor styles TopBar and TopBarPageTitle

Cray-code:


src/components/TopBar/TopBarPageTitle.js
```diff
@@ -14,10 +14,7 @@ export const TopBarPageTitle = ({ children, ...props }) => {
 
   return (
     <div className={classes.topBarPageTitle}>
-      <Typography
-        variant="h6"
-        style={{ color: 'rgba(255, 255, 255, 0.87)' }}
-        {...props}>
+      <Typography variant="h6" style={{ color: colors.white87 }} {...props}>
```
DarthVanger, line 17: I think we could add this style to [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L16), where we have all the other styles for `h6`.

src/components/TopBar/TopBar.js
```diff
@@ -12,7 +13,7 @@ const TopBarWrapper = styled.div`
 export const TopBar = ({ children }) => {
   return (
     <TopBarWrapper>
-      <AppBar style={{ background: '#272727' }}>
+      <AppBar style={{ background: colors.elevationOverlay04dp }}>
```
DarthVanger, line 16: We already have the background color for `Appbar` defined in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L64):
```
  overrides: {
    MuiAppBar: {
      colorPrimary: {
        backgroundColor: null,
        background: colors.elevationOverlay04dp,
      },
    },
 ```
 
 Maybe we can just delete the style here and it will use the color from the theme?

Comment (Cray-code) [reply src/components/TopBar/TopBar.js:16]: Yes, it use the color from the theme.

File: src/components/TopBar/TopBarPageTitle.js (hunk 1 above)
Comment (Cray-code) [reply src/components/TopBar/TopBarPageTitle.js:17]: Add ```typography: title1``` style to theme.js

File: src/theme.js [hunk 3]
```diff
@@ -38,6 +38,15 @@ export const theme = createMuiTheme({
       letterSpacing: 0.1,
       color: 'rgba(255, 255, 255, 0.6)',
     },
+    h6: {
```
DarthVanger, line 41: Is `headline6` and `h6` two different things, or maybe it's the same?
We already have some styles for `headline6` in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L16):
```
    headline6: {
      fontFamily: 'Roboto',
      fontStyle: 'normal',
      fontWeight: 500,
      fontSize: '19px',
      lineHeight: '22px',
    },
  ```

DarthVanger (approved):

PR #216, opened 2021-04-20
Note details page

kadokan81:

PR #218, opened 2021-04-20
Fixsearch

kadokan81:


src/components/BottomBar/BottomBarSagas.js
```diff
@@ -25,6 +25,7 @@ function* handleChangeParentButtonClick({ data: { note } }) {
 function* deleteNote({ data: { note } }) {
   yield noteStorage.remove(note);
   yield put(deleteNoteRequestSuccessAction(note));
+  yield put(push('/notes'));
```
DarthVanger, line 28: We don't need this redirect in `BottomBarSagas.js`.

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+  title: {
+    color: colors.white87,
+  },
+}));
+
+const Transition = React.forwardRef(function Transition(props, ref) {
+  return <Slide direction="up" ref={ref} {...props} />;
+});
+
+export default function ModalDeleteComponent({ handleClose, open }) {
```
DarthVanger, line 28: Let's rename this component to `DeleteNoteConfirmationDialog`.

"ModalDeleteComponent" isn't very clear as to what is this component for and about.

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+  const selectedNote = useSelector(state => state.notesMindMap.selectedNote);
+  const dispatch = useDispatch();
+  const handleDeleteButtonClick = () => {
+    dispatch(deleteNoteAction(selectedNote));
+  };
+  const classes = useStyles();
+  return (
+    <div>
```
DarthVanger, line 36: I think we can remove this `<div>`. Do we really need it?

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+      <Dialog
+        open={open}
+        TransitionComponent={Transition}
+        keepMounted
+        onClose={handleClose}
+        aria-labelledby="alert-dialog-slide-title"
+        aria-describedby="alert-dialog-slide-description">
+        <DialogTitle className={classes.dialog} id="alert-dialog-slide-title">
+          {'Delete None'}
```
DarthVanger, line 45: As per [design](https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836), we should show `Delete {note.name}` here:
image

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+        </DialogTitle>
+        <DialogContent className={classes.dialog}>
+          <DialogContentText
+            className={classes.title}
```
DarthVanger, line 49: Why does `DialogContentText` have className `classes.title` ?
I assume `classes.title` is for `<DialogTitle>` and not for `DialogContentText` ?

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+import DialogTitle from '@material-ui/core/DialogTitle';
+import Slide from '@material-ui/core/Slide';
+import { useDispatch, useSelector } from 'react-redux';
+import { deleteNoteAction } from 'components/BottomBar/BottomBarActions';
+import { makeStyles } from '@material-ui/core';
+import { colors } from '../../colors';
+
+const useStyles = makeStyles(() => ({
+  dialog: {
+    color: 'white',
```
DarthVanger, line 16: Please use `<Typography variant="body2">` for dialog text style:

image

We already have styles defined for this text in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L43).

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+    backgroundColor: colors.mainGray,
```
DarthVanger, line 17: Shouldn't we use `colors.dialogsGray` instead of `colors.mainGray` here?

image

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+  },
+  title: {
+    color: colors.white87,
```
DarthVanger, line 20: Please use `<Typography variant="subtitle1">` for the this text:

image

We have already defined styles for this text in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L16)

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <div>
+      <Dialog
+        open={open}
+        TransitionComponent={Transition}
+        keepMounted
+        onClose={handleClose}
+        aria-labelledby="alert-dialog-slide-title"
+        aria-describedby="alert-dialog-slide-description">
+        <DialogTitle className={classes.dialog} id="alert-dialog-slide-title">
```
DarthVanger, line 44: Why does `<DialogTitle>` has className `classes.dialog` ?

Shouldn't it be `classes.title` ?

DarthVanger, line 44: The `id` should be unique across the whole html document.
"alert-dialog-slide-title" doesn't sound like a unique name: if we add more dialogs they might also have the same id, copied from the Material UI docs :)
Since you're using `keepMounted` prop, this component will always be kept in the DOM.
So we have a real chance of having two elements with the same `id` in the DOM, preventing the `aria-labelledby` from finding the correct element in the DOM (it will take the first element it found by the id).

Let's give it a more unique id, like `"delete-note-confirmation-dialog-title"`.
This way if we add another dialogue in the future, it will most probably have a different `id`.

Same for the `alert-dialog-slide-description` id.

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+          {'Delete None'}
+        </DialogTitle>
+        <DialogContent className={classes.dialog}>
+          <DialogContentText
+            className={classes.title}
+            id="alert-dialog-slide-description">
+            Are you sure you want to delete this note?
+          </DialogContentText>
+        </DialogContent>
+        <DialogActions className={classes.dialog}>
```
DarthVanger, line 54: Why `<DialogActions>` have className `classes.dialog` ?
I don't think this className is needed here at all, as we don't use the `white` color in the dialog actions. The buttons use `primary` color.

src/components/ModalDelete/ModalDeleteComponent.js
```diff
@@ -0,0 +1,69 @@
...
+          <Button onClick={handleClose} color="primary">
+            Cancel
+          </Button>
+          <Button
+            onClick={() => {
+              handleDeleteButtonClick(), handleClose();
```
DarthVanger, line 60: I think it would be a bit cleaner if have
```
  <Button onClick={handleDeleteButtonClick}>
```
and them in the `handleDeleteButtonClick` function
```
const handleDeleteButtonClick = () => {
      dispatch(deleteNoteAction(selectedNote));
      handleClose();
}
```

src/components/NoteDetails/NoteDetailsPage.js
```diff
@@ -1,31 +1,33 @@
...
 
-const BackButtonWrapper = styled.div`
-  color: red !important;
-  padding: 1rem;
+const Wrapper = styled.div`
+  background-color: ${colors.darkViolet};
 `;
 
 export const NoteDetailsPage = () => {
+  const [open, setOpen] = useState(false);
```
DarthVanger, line 17: It is ambiguous what does `open` mean: since it's in the `NoteDetailsPage` I would assume it means that the NoteDetailsPage is either open or closed.

Let's name it `isDeleteNoteConfirmationDialogOpen`: this way it would be clear what is this state variable used for.

src/components/NoteDetails/NoteDetailsPage.js
```diff
@@ -1,31 +1,33 @@
...
     <Wrapper>
-      <BackButtonWrapper>
-        <Link to="/notes">
-          <IconButton aria-label="back">
-            <ArrowBackIcon style={{ fill: '#E4E3E4' }} />
-          </IconButton>
-        </Link>
-      </BackButtonWrapper>
+      <NoteDetailsTopBar handleClickOpen={handleClickOpen} />
+      <ModalDeleteComponent open={open} handleClose={handleClose} />
```
DarthVanger, line 30: I think it would be better to put this modal into `NoteDetailsTopBar` component, since the button opening the modal is there.
And also `changeParent` button click is handled inside `NoteDetailsTopBar`.
I believe it would make more sense to handle all the buttons in `NoteDetailsTopBar`, rather than handling one button in `NoteDetailsTopBar` and another button in `NoteDetailsPage`.

src/components/NoteDetails/NoteDetailsPage.js
```diff
@@ -1,31 +1,33 @@
...
   return (
     <Wrapper>
-      <BackButtonWrapper>
-        <Link to="/notes">
-          <IconButton aria-label="back">
-            <ArrowBackIcon style={{ fill: '#E4E3E4' }} />
-          </IconButton>
-        </Link>
-      </BackButtonWrapper>
+      <NoteDetailsTopBar handleClickOpen={handleClickOpen} />
```
DarthVanger, line 29: `handleClickOpen` sounds quite confusing, it's hard to understand what is it for by reading the prop name.
Let's rename it to `handleDeleteButtonClick`

src/components/NoteDetails/NoteDetailsTopBar.js
```diff
@@ -0,0 +1,98 @@
+import React from 'react';
```
DarthVanger, line 1: This code is the same as in another pull request: https://github.com/DarthVanger/NeuralNotes/pull/216/files.

This code is not for search and shouldn't be in this pull request. Please remove it.

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+  customInput: {
+    color: colors.iconColor,
+    padding: '0px 10px',
+  },
+  list: {
+    width: '100%',
+    height: '100%',
+    paddingLeft: 16,
+    marginTop: -30,
```
DarthVanger, line 25: Why do we need a negative margin here?

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+  },
+  listItem: {
+    padding: 0,
+    height: 64,
+  },
+  icon: {
+    paddingLeft: 12,
+    paddingRight: 12,
+  },
+  svgIcon: {
```
DarthVanger, line 35: What is the difference between `icon` and `svgIcon` ? 
Aren't all icons in svg format?

I think we can rename `icon` to `listItemIcon` and `svgIcon` to `icon`.

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+import { searchQueryChangeAction } from './SearchPageAction';
+import { useEffect } from 'react';
+import { colors } from 'colors';
+import ClearIcon from '@material-ui/icons/Clear';
+
+const useStyles = makeStyles(() => ({
+  multilineColor: {
+    color: colors.iconColor,
+  },
+  customInput: {
```
DarthVanger, line 17: I think this should be named `textField`, as these styles are applied to the `<TextField>` component.

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+import { IconButton, makeStyles } from '@material-ui/core';
+import { useDispatch } from 'react-redux';
+import { useDebouncedCallback } from 'use-debounce/lib';
+import { searchQueryChangeAction } from './SearchPageAction';
+import { useEffect } from 'react';
+import { colors } from 'colors';
+import ClearIcon from '@material-ui/icons/Clear';
+
+const useStyles = makeStyles(() => ({
+  multilineColor: {
```
DarthVanger, line 14: What do you mean by `multiline` here?
I think this style should be named `input` and it's applied to the input.

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+    width: '100%',
+    height: '100%',
+    paddingLeft: 16,
+    marginTop: -30,
+  },
+  listItem: {
+    padding: 0,
+    height: 64,
+  },
+  icon: {
```
DarthVanger, line 31: It looks like this style is not used in this file.

src/components/SearchPage/SearchInputComponent.js
```diff
@@ -0,0 +1,84 @@
...
+        },
+        startAdornment: (
+          <InputAdornment position="start">
+            <SearchIcon style={{ fill: colors.white60 }} />
+          </InputAdornment>
+        ),
+
+        endAdornment: value && (
+          <IconButton
+            aria-label="toggle password visibility"
```
DarthVanger, line 76: We don't have any passwords in this input :)
It should be something like "clear search".

src/components/SearchPage/SearchInputComponent.js
kadokan81: Just copy paste from docs.  renamed.

src/components/SearchPage/SearchInputComponent.js
kadokan81: Renamed.

src/components/SearchPage/SearchInputComponent.js
kadokan81: Definitely no, deleted

src/components/SearchPage/SearchInputComponent.js
kadokan81: Deleted

src/components/SearchPage/SearchInputComponent.js
kadokan81: deleted

PR #219, opened 2021-04-21
Show "no results found" when search returns nothing

SerhiiPodaiko:


src/components/NotFound/NotFoundComponent.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { StyledNotFound } from 'components/NotFound/NotFoundStyles';
+
+export function NotFound() {
+  return (
+    <StyledNotFound>
+      <div className="not-found">
+        {/* eslint-disable-next-line react/no-unescaped-entities */}
+        <h3>We didn't find any matches</h3>
```
DarthVanger, line 9: With Material UI we should use `<Typography variant="h3">` instead of a usual `<h3>`, see https://material-ui.com/components/typography/

In our [design](https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836) though we don't have `h3`, but we have `h6`:
image

We have the styles for it defined in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L34)

So I think it would be best to use `<Typography variant="h6">` here, as we already have proper styles for it: color, font family, font weight, letter spacing, etc.

Issue #194, opened 2020-11-22
Implement TopBar for Search screen


DarthVanger:
Currently the SearchPage has no Topbar:

![image](https://user-images.githubusercontent.com/6686679/99906478-11ed6880-2ce0-11eb-846b-1d319ebac485.png)

Implement a topbar there as per design: https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836

Also there is a weird code for [BackButtonWrapper](https://github.com/DarthVanger/NeuralNotes/blob/6da6d7765baf3cb82f8f540bace2db93abd01df6/src/components/SearchPage/SearchPageComponent.js#L23) with `color: red !important;`. Remove `!important` and I don't think we're using `red` color anywhere, so I guess the whole thing should be removed.

And there is a [hardcoded color for the back button arrow](https://github.com/DarthVanger/NeuralNotes/pull/167/files#diff-2013711681cb591e532b0cf771139ca02a102ff29f54228ae7193aab39ca1e9fR42). We should use color from `colors.js` instead or from material-ui theme.
```
     <ArrowBackIcon style={{ fill: '#BB86FC' }} />
```

DarthVanger:
Implemented in https://github.com/DarthVanger/NeuralNotes/pull/231

PR #220, opened 2021-04-21
Show not found message

SerhiiPodaiko:
Show not found message for search result

### DIFF
--- a/src/components/NotFound/NotFoundComponent.js
+++ b/src/components/NotFound/NotFoundComponent.js
@@ -1,0 +1,14 @@
+import React from 'react';
+import { StyledNotFound } from 'components/NotFound/NotFoundStyles';
+
+export function NotFound() {
+  return (
+    <StyledNotFound>
+      <div className="not-found">
+        {/* eslint-disable-next-line react/no-unescaped-entities */}
+        <h3>We didn't find any matches</h3>
+        <span>Please check your input data</span>
+      </div>
+    </StyledNotFound>
+  );
+}
--- a/src/components/NotFound/NotFoundStyles.js
+++ b/src/components/NotFound/NotFoundStyles.js
@@ -1,0 +1,12 @@
+import styled from 'styled-components';
+import { colors } from 'colors';
+
+export const StyledNotFound = styled.div`
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  color: red;
+  background: ${colors.darkGray};
+  height: ${'calc(100% - 101px)'};
+`;
--- a/src/components/SearchList/SearchList.js
+++ b/src/components/SearchList/SearchList.js
@@ -1,0 +1,47 @@
+import React from 'react';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemIcon from '@material-ui/core/ListItemIcon';
+import InsertDriveFileRoundedIcon from '@material-ui/icons/InsertDriveFileRounded';
+import ListItemText from '@material-ui/core/ListItemText';
+import { Typography } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import { colors } from 'colors';
+
+const useStyles = makeStyles(() => ({
+  list: {
+    width: '100%',
+    height: 'calc(100% - 101px)',
+    paddingLeft: 16,
+  },
+  listItem: {
+    padding: 0,
+    height: 64,
+  },
+  icon: {
+    paddingLeft: 12,
+    paddingRight: 12,
+  },
+  svgIcon: {
+    color: colors.iconColor,
+  },
+}));
+
+export const SearchList = ({ resultList, handleItemClick }) => {
+  const styleClasses = useStyles();
+
+  return (
+    <List className={styleClasses.list}>
+      {resultList.map(searchResult => (
+        <ListItem key={searchResult.id} className={styleClasses.listItem}>
+          <ListItemIcon className={styleClasses.icon}>
+            <InsertDriveFileRoundedIcon className={styleClasses.svgIcon} />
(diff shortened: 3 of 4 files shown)


src/components/NotFound/NotFoundComponent.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { StyledNotFound } from 'components/NotFound/NotFoundStyles';
+
+export function NotFound() {
+  return (
+    <StyledNotFound>
+      <div className="not-found">
+        {/* eslint-disable-next-line react/no-unescaped-entities */}
```
DarthVanger, line 8: Ignoring eslint here doesn't look like a great solution :)
Let's better do what they recommend on stackoverflow :)
https://stackoverflow.com/a/45898131/1657101

The same advice is given on the [eslint rule page](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md):
image

Or the following might work as well:
```
<h3>{'We didn\'t find any matches'}</h3>
```

src/components/SearchList/SearchList.js
```diff
@@ -0,0 +1,47 @@
...
+import InsertDriveFileRoundedIcon from '@material-ui/icons/InsertDriveFileRounded';
+import ListItemText from '@material-ui/core/ListItemText';
+import { Typography } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import { colors } from 'colors';
+
+const useStyles = makeStyles(() => ({
+  list: {
+    width: '100%',
+    height: 'calc(100% - 101px)',
```
DarthVanger, line 14: Looks weird... why `101px` ? :)

src/components/SearchPage/SearchPageComponent.js
```diff
@@ -1,51 +1,33 @@
...
-  icon: {
-    paddingLeft: 12,
-    paddingRight: 12,
-  },
-  svgIcon: {
-    color: colors.iconColor,
-  },
-}));
+const Container = styled.div`
+  color: red !important;
```
DarthVanger, line 20: Let's not use `important`. Do we even need a `red` color here?

src/components/SearchPage/SearchPageComponent.js
```diff
@@ -1,51 +1,33 @@
...
+  width: 100%;
+  height: 100%;
+`;
+
+const InputWrapper = styled.div`
+  padding-left: 20px;
```
DarthVanger, line 26: The Search page topbar styles are being changed in https://github.com/DarthVanger/NeuralNotes/issues/194 by @kadokan81 .
I think it's better to avoid changing any styles for search input or back button in this PR in order to prevent conflicts with the [PR](https://github.com/DarthVanger/NeuralNotes/issues/194) created by @kadokan81 for the Search page TopBar.

src/components/NotFound/NotFoundStyles.js
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components';
+import { colors } from 'colors';
+
+export const StyledNotFound = styled.div`
+  width: 100%;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  color: red;
```
DarthVanger, line 9: Why `red`? I don't think we have `red` color anywhere in our design.
Let's use one of the `<Typography>` variants for the text style. You can see which variants we have defined in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L15).

For example in Google Drive they don't use `red`. They use the same text style as for the other texts on their website:
image

src/components/NotFound/NotFoundComponent.js
```diff
@@ -0,0 +1,14 @@
...
+        <h3>We didn't find any matches</h3>
```
DarthVanger, line 9: With Material UI we should use `<Typography variant="h3">` instead of a usual `<h3>`, see https://material-ui.com/components/typography/

In our [design](https://www.figma.com/file/FxmwqO6LRg7tOVXrZqMMG0/Neural-Notes-Design-v.0.3?node-id=212%3A836) though we don't have `h3`, but we have `h6`:
image

We have the styles for it defined in [theme.js](https://github.com/DarthVanger/NeuralNotes/blob/master/src/theme.js#L34)

So I think it would be best to use `<Typography variant="h6">` here, as we already have proper styles for it: color, font family, font weight, letter spacing, etc.

src/components/NotFound/NotFoundComponent.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import { StyledNotFound } from 'components/NotFound/NotFoundStyles';
+
+export function NotFound() {
+  return (
+    <StyledNotFound>
+      <div className="not-found">
```
DarthVanger, line 7: Do we need this `className`? Is it used for anything?

src/components/NotFound/NotFoundStyles.js
```diff
@@ -0,0 +1,12 @@
...
+  background: ${colors.darkGray};
+  height: ${'calc(100% - 101px)'};
```
DarthVanger, line 11: Looks weird... why 101px ? :)

src/components/SearchPage/SearchPageComponent.js
```diff
@@ -64,39 +46,29 @@ export function SearchPageComponent() {
     dispatch(searchResultClickedAction({ note }));
   };
 
-  const classes = useStyles();
+  const responseList =
```
DarthVanger, line 49: Why `responseList` if it's JSX for search results? :)
Since you have renamed `searchResults` into `resultList`, maybe you can name this variable `resultListElement`.
Or make it a function `renderResultList()`.

DarthVanger:
Closing due to inactivity

PR #232, opened 2021-04-23
create NoteDetailsTopBar element , add padding-top to StyledNoteDetai…

kadokan81:
create NoteDetailsTopBar element , add padding-top to StyledNoteDetaisScreen, implement top menu dropdown

### DIFF
--- a/src/components/NoteDetails/NoteDetailsStyles.js
+++ b/src/components/NoteDetails/NoteDetailsStyles.js
@@ -6,7 +6,7 @@ export const StyledNoteDetailsScreen = styled.div`
   flex-grow: 1;
   display: flex;
   flex-direction: column;
-  padding-top: 70px;
+  padding-top: 60px;
 `;
 
 export const DefaultTextarea = styled.textarea`
@@ -18,7 +18,7 @@ export const DefaultTextarea = styled.textarea`
   border-radius: 4px;
   outline: none;
   box-shadow: none;
-  margin-bottom: 20px;
+  margin-bottom: 16px;
 `;
 
 export const StyledNoteNameEditor = styled(DefaultTextarea)`
--- a/src/components/NoteDetails/NoteDetailsStyles.js
+++ b/src/components/NoteDetails/NoteDetailsStyles.js
@@ -18,7 +18,6 @@ export const DefaultTextarea = styled.textarea`
   border-radius: 4px;
   outline: none;
   box-shadow: none;
-  margin-bottom: 16px;
 `;
 
 export const StyledNoteNameEditor = styled(DefaultTextarea)`
--- a/src/components/NoteDetails/NoteDetailsStyles.js
+++ b/src/components/NoteDetails/NoteDetailsStyles.js
@@ -27,6 +27,7 @@ export const StyledNoteNameEditor = styled(DefaultTextarea)`
   font-size: 20px;
   font-weight: 500;
   flex-grow: 0;
+  margin-top: -24px; // override status message 'Saved...'
 `;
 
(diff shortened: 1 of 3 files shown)


src/components/NoteDetails/NoteDetailsTopBar.js
```diff
@@ -0,0 +1,88 @@
...
+import { useDispatch, useSelector } from 'react-redux';
+import { useRef, useState } from 'react';
+import { TopBarPageTitle } from 'components/TopBar/TopBarPageTitle';
+import { changeParentButtonClickedAction } from 'components/BottomBar/BottomBarActions';
+import { colors } from 'colors';
+import { BackButton } from 'components/BackButton/BackButton';
+
+const useStyles = makeStyles(() => ({
+  moreIcon: {
+    position: 'absolute',
```
DarthVanger, line 21: Did you try using `TopBarRightButtons` component? I think it already has styles to position the buttons correctly on the right side of the topbar.

src/components/NoteDetails/NoteDetailsTopBar.js
```diff
@@ -0,0 +1,88 @@
...
+    color: colors.textColor,
+    right: 0,
+    padding: '12px',
+    marginRight: '40px',
```
DarthVanger, line 25: Looks like this margin is too big, compared to what we have in the design.
If you use `TopBarRightButtons` component I think it should apply correct styles.

src/components/NoteDetails/NoteDetailsTopBar.js
```diff
@@ -0,0 +1,88 @@
...
+  const dispatch = useDispatch();
+
+  const openMenu = () => setIsMenuOpened(true);
+  const closeMenu = () => setIsMenuOpened(false);
+
+  const handleChangeParentButtonClick = () => {
+    dispatch(changeParentButtonClickedAction(selectedNote));
+  };
+
+  const title = selectedNote ? selectedNote.name : 'Add you note';
```
DarthVanger, line 43: I believe we always have a `selecteNote` in Redux. So `title` will never be `'Add you note'`.
I tried creating a new note on your branch and it shows the name of `selectedNote`:

image

I will create a separate task to fix the title problem when creating a new note.
For now let's just show the `selectedNote.name` as the topbar title.

src/components/NoteDetails/NoteDetailsStyles.js
```diff
@@ -17,6 +18,7 @@ export const DefaultTextarea = styled.textarea`
   border-radius: 4px;
   outline: none;
   box-shadow: none;
+  margin-bottom: 20px;
```
DarthVanger, line 21: In the design the distance between the note name and note content is `16px`:

image

Why did you add a `20px` margin ?

DarthVanger: And in this task you're working on the topbar styles. Why do you even need to change the `textarea` styles?

src/components/NoteDetails/NoteDetailsStyles.js
```diff
@@ -26,7 +27,6 @@ export const StyledNoteNameEditor = styled(DefaultTextarea)`
   font-size: 20px;
   font-weight: 500;
   flex-grow: 0;
-  margin-top: -24px; // override status message 'Saved...'
```
DarthVanger, line 29: Please add this line back, as it breaks the styles:
Before your change:
image

After your change:
image

Notice a black block between the topbar and the note name editor

DarthVanger (approved):

PR #235, opened 2021-04-25
197 note editor http errors

Cray-code:
The functionality is mostly implemented, it needs discussion...

### DIFF
--- a/src/components/NoteDetails/NoteDetailsComponent.js
+++ b/src/components/NoteDetails/NoteDetailsComponent.js
@@ -1,12 +1,12 @@
 import React, { useState, useEffect } from 'react';
 import { useDebounce } from 'use-debounce';
-import { Typography } from '@material-ui/core';
 
 import {
   StyledNoteNameEditor,
   StyledNoteDetailsScreen,
   StyledNoteContentEditor,
-} from 'components/NoteDetails/NoteDetailsStyles';
+} from './NoteDetailsStyles';
+import NoteDetailsSaveStatus from './NoteDetailsSaveStatus';
 
 export const NoteDetailsComponent = props => {
   const [noteName, setNoteName] = useState(props.noteName);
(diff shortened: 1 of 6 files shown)


src/components/NoteDetails/NoteDetailsSagas.js
```diff
@@ -145,29 +147,57 @@ function* applyQueuedNoteUpdate({ data: { note, queuedChanges: changes } }) {
...
+    console.error(e);
   }
 }
 
 function* createNoteRequest({ data: { note } }) {
-  const newNote = yield noteStorage.create(note);
-  newNote.parent = note.parent;
-  yield put(createNoteSuccessAction(newNote));
+  try {
+    const newNote = yield noteStorage.create(note); // func return object (not Promise) or error
```
DarthVanger, line 188: If it doesn't return promise, why are we using `yield`? 
I believe it does return a promise, because it's an asynchronous operation.

src/components/NoteDetails/NoteDetailsSnackbar.js
```diff
@@ -0,0 +1,59 @@
...
+    if (reason === 'clickaway') {
+      return;
+    }
+
+    setOpen(false);
+
+    if (error.text.includes('name')) {
+      saveNoteName();
+    }
+    if (error.text.includes('content')) {
```
DarthVanger, line 28: Having logic based around `error.text` includes `'something'` doesn't look like a good idea.
For example a better approach would be to have
```
errors = {
  saveNoteName: ...,
  saveNoteContent: ...,
  createNote: ...,
  };
  ```
  Then here we would just do
  ```
  if (errors.saveNoteContent) {
     ...
  }
  ```
Using JS variables is more robust and clear, rather than relying on the error text string having some word included into it.

File: src/components/NoteDetails/NoteDetailsComponent.js [hunk 3]
```diff
@@ -1,12 +1,13 @@
...
-import { Typography } from '@material-ui/core';
 
 import {
   StyledNoteNameEditor,
   StyledNoteDetailsScreen,
   StyledNoteContentEditor,
-} from 'components/NoteDetails/NoteDetailsStyles';
+} from './NoteDetailsStyles';
+import NoteDetailsSaveStatus from './NoteDetailsSaveStatus';
+import NoteDetailsSnackbar from './NoteDetailsSnackbar';
```
DarthVanger, line 10: I would name it `NoteSaveFailedSnackbar` or `NoteSaveFailedNotification`

DarthVanger:
Closing due to inactivity

